PR #108, opened 2018-02-17
A few bug fixes

SkippyleGrandGourou:
Here are a few minor bugfixes, which should be straightforward to understand from both code and comments, but obviously I can elaborate if needed.

One point about the empty search error fix : it seems from the "*query*" that the intent was to match any column which contains "query".  However it appears this is not the way MATCH AGAINST works, see [MySQL documentation](https://dev.mysql.com/doc/refman/5.6/en/fulltext-boolean.html).  The fix assumes that the intent is to search for a full word.

### DIFF
--- a/public_html/includes/templates/default.catalog/pages/printable_order_copy.inc.php
+++ b/public_html/includes/templates/default.catalog/pages/printable_order_copy.inc.php
@@ -241,7 +241,16 @@ hr {
         <div class="value"><?php echo settings::get('store_email'); ?></div>
 
         <div class="label"><?php echo language::translate('title_website', 'Website'); ?></div>
-        <div class="value"><?php echo htmlspecialchars(document::ilink('')); ?></div>
+        <div class="value">
+          <?php
+            $url = document::href_ilink('');
+            if (substr($url,0,5) == "https") {
+              echo "https://" . idn_to_utf8(substr($url,8));
+            } else {
+              echo "http://" . idn_to_utf8(substr($url,7));
+            }
+          ?>
+        </div>
       </div>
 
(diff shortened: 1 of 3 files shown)


timint:
Thanks for creating a pull request. For the commit "Taking care of internationalized domain names" I'm wondering if maybe IDN decoding should be added to a central point like in compatibility.inc.php or lib_link.inc.php instead.

I've been going back and fourth with the product groups thing wheither it should be visible or hidden when there is only one option listed. What's your input on this?

timint:
The native idn decode is utf8 only. Something like this would be necessary:

```
function decode_idn($url) {

      if (!function_exists('idn_to_utf8')) return $url;

      if (substr($url, 0, 8) == 'https://') {
        $url = 'https://' . idn_to_utf8(substr($url, 8));

      } else if (substr($url, 0, 7) == 'http://') {
        $url = 'https://' . idn_to_utf8(substr($url, 8));

      } else {
        $url = idn_to_utf8($url);
      }

      return language::convert_characters($url, 'UTF-8');
    }
```

SkippyleGrandGourou:
I'm afraid I'm not experienced enough with litecart (and e-commerce platforms in general) to have an opinion on where IDN should be treated. As far as I'm concerned, as long as the IDN is readible by the customer, I'm happy.

Concerning the product groups : since product groups are independant (you don't choose between product groups but between groups within a product group), I don't see why it shouldn't appear if there is only one. Actually, it is quite confusing that nothing happens on the store frontend when you create a product group and populate it.

timint:
I have committed most of your changes, slightly justifiied. I'm gonna close this pull request as they are parallelly commited. Thank you for providing solutions.

PR #121, opened 2018-05-16
Fix: Undefined variable

cbekir:


litecart:
Thank you for providing a fix.
I am closing this pull request as it was targeting the master branch instead of dev (which is where the commits go). The fix is now committed to the dev branch.

PR #150, opened 2019-04-09
set_time_limit()  causes a warning on shared hosting

litecart-cs:
set_time_limit() has been disabled for security reasons
mostly on shared hosting


timint:
Very good, I did nail that in push_jobs.inc.php but not in modules.app/run_job.inc.php. I have too much in my commit list for merging your pull request. Therefore I will make these commits manually and close this pull request. Thank you for submitting. :)

PR #152, opened 2019-05-15
Missing translate

litecart-cs:


timint:
It's not missing, It's on row 28. I think you mean
`      $this->name = language::translate(__CLASS__.':title', 'Cash on Delivery');`

PR #157, opened 2019-07-12
inconsistent product categories save

SkipperHosting:
product categories would not consistently save via admin 'edit product' app - only primary category would stay set, and could only 'add' cats, not remove from product without problems.  Follows convention of other similar checks to suppress error with implode.


timint:
Thank you for fixing. Will be merged. :)

PR #158, opened 2019-07-13
Sets first radio as checked

SkipperHosting:
Currently, a product's radio button group is drawn with no element 'checked'.

This sets the first option in the group as checked.  Thus, providing a more proper experience for the customer.  Respects any already checked radio button in the group.  As well, updates page on load to account for any price changing options that may need to show on the final price.


SkipperHosting:
Sorry for all the extraneous commits, still learning git

timint:
Will cause problems with optional choices. I'm looking into rewriting how options and stocks are handled.

PR #165, opened 2019-10-23
Update ent_product.inc.php

Nicolasmol:
Adding a condition, to avoid cumuling in error.log messages about non existing images to be renamed

### DIFF
--- a/public_html/includes/entities/ent_product.inc.php
+++ b/public_html/includes/entities/ent_product.inc.php
@@ -531,10 +531,10 @@
             $this->data['images'][$key]['id'] = database::insert_id();
           }
 
-          if (!empty($this->data['images'][$key]['new_filename'])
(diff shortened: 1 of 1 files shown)


timint:
Thanks for submitting a pull request. I am denying it because if an image is missing in the first place that is surely a sign of an error which the merchant should be aware of. And the indentation of your commit seemed wrong viewing it from an ios device. Other than that happy to recceive pull requests and see that ppl have ideas. :)

Nicolasmol:
1) about the «  sign of error »... not really, because this occurs for example, in my case, after having CSV imported the products, and before which user can’t be perfect in forecasting the litecart’s pictures naming rules.
2) about indentation : it’s always a problem since, I surely don’t use the most commun rules (tab/spaces & number) 

Nicolas Molinari
Envoyé de mon iPhone

PR #168, opened 2019-10-30
Update func_catalog.inc.php

Nicolasmol:
Restoring display of products when in "catalog only mode"

### DIFF
--- a/public_html/includes/functions/func_catalog.inc.php
+++ b/public_html/includes/functions/func_catalog.inc.php
@@ -253,7 +253,9 @@
       from (
         select p.id, p.sold_out_status_id, p.code, p.sku, p.mpn, p.gtin, p.manufacturer_id, group_concat(ptc.category_id separator ',') as categories, p.keywords, p.image, p.tax_class_id, p.quantity, p.views, p.purchases, p.date_created
         from ". DB_TABLE_PRODUCTS ." p
-        left join ". DB_TABLE_PRODUCTS_TO_CATEGORIES ." ptc on (p.id = ptc.product_id)
+        left join ". DB_TABLE_PRODUCTS_TO_CATEGORIES ." ptc on (p.id = ptc.product_id)" .
+        (
+        settings::get('catalog_only_mode') ? "" : "
         left join ". DB_TABLE_SOLD_OUT_STATUSES ." ss on (p.sold_out_status_id = ss.id)
         where p.status
           and (p.id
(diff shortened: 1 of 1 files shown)


litecart:
Your commit removes the filtering of products if the catalog only mode is enabled.

Nicolasmol:
But as far I’m reading the MySQL request 
How can be considered as being necessarily important not to auto-hide products based on 
- quantity stock = 0
or
- sold-out-status = hidden

Since ... A catalog mode only doesn’t sale nothing !!!!?


Nicolas Molinari
Envoyé de mon iPhone

litecart:
If you have products in stock, or set a sold out status NOT to hide them. They should be visible.

Catalog only mode just turns off the Add to Cart button and Checkout.

`and (p.quantity > 0 or ss.hidden != 1)`

Nicolasmol:
Hum 
Usually ppl who turns their shop from sailing one to non-saling one usually faces a sad story in their business life  and I hope this is not the most usual case, which I think is this opposite : from non-saling to saling (or stay non-saling for ever)...

Nicolas Molinari
Envoyé de mon iPhone

litecart:
I'm not sure what that would have to do with stock or sold out statuses. And making a statement like that it sounds like you would speak for a whole crows of merchants. A reference might be good.

Nicolasmol:
Let’s take this problem the opposite side then.
How can keep displaying the similar products box as it is in 2.16 when using in catalog mode only ?
Is it then necessary to set a virtual quantity to each product?

Nicolas Molinari
Envoyé de mon iPhone

litecart:
Catalog only mode should not have any effect like that. The sold out status would.
2.1.6 did not have the ability to hide sold out products.

Nicolasmol:
Maybe better like this:



and (p.quantity > 0 or ss.hidden == 0)



instead of :



and (p.quantity > 0 or ss.hidden != 1)


since ss.hidden  == null


when ppl don't care fill "stock" things


Le 1 novembre 2019 à 11:54, LiteCart <notifications@github.com> a écrit :



Catalog only mode should not have any effect like that. The sold out status would.
2.1.6 did not have the ability to hide sold out products.

Nicolasmol:
I mean (I forgot I was inside a Mysql reques)



and (p.quantity > 0 or ss.hidden != 1 or isnull(ss.hidden))

Le 1 novembre 2019 à 11:54, LiteCart <notifications@github.com> a écrit :



Catalog only mode should not have any effect like that. The sold out status would.
2.1.6 did not have the ability to hide sold out products.

litecart:
LiteCart tables today does not use NULL.

Nicolasmol:
However works

Nicolas Molinari
Envoyé de mon iPhone

PR #170, opened 2019-12-15
`date_login` can be NULL

litecart-cs:
"NOT NULL" for `date_login` is too strict. 
Importing customers from another/accounting/CRM system and so on... = problem.


timint:
LiteCart does not support NULL values because en empty string and null is not the same thing so every query condition would have to be rewritten to support both.

where (column == '' or column is null)

Disable MySQL Strict and you should be fine.

PR #190, opened 2020-03-25
Avoid autoload error (composer issue)

arshidkv12:


timint:
Composer should prepend the spl_autoload_register(). Does it not?

See prepend:
https://www.php.net/spl_autoload_register

arshidkv12:
It is making trouble, if we create addon with composer. So I added
file_exists function to avoid this problem.

timint:
Is it so that Composer prepend autoload is set to true but the composer package it tries to access is missing? There are several ways to solve this. And I wanna be sure this is it.

arshidkv12:
I added stripe library with composer in module folder. Then I included
autoload.php in my module. Then it is making error.

timint:
What happens if you remove your autoloader and just rely on this?
https://github.com/litecart/litecart/blob/dev/public_html/includes/app_header.inc.php#L26

arshidkv12:
Not loading vendor autoload file inside module directory (eg: payment)

timint:
And your vendor autoload file is ~/vendors/autoload.php?
And you say it's Composer.

It makes no sense. Composer should always prepend the autoloader as included in includes/app_header.inc.php.

arshidkv12:
I separately downloaded stripe files by composer. My vendor folder is
inside include/module/payment/stripe/vendor.

timint:
See if this is a better solution
https://www.litecart.net/en/addons/255/composer-for-litecart

arshidkv12:
Please note, I am developing addon.

arshidkv12:
Please check responsive issue also  https://prnt.sc/rp9pkd

timint:
Closing as conflicting code and seems to be a few versions behind.

arshidkv12:
Cloned from dev branch.

PR #241, opened 2020-09-05
Update lib_vmod.inc.php

arshidkv12:
Class VQMod referenced with incorrect case: vqmod.


timint:
That's a better pull request. :)
Did PHP give you any trouble for this? class names should not be case sensitive but I am unsure about PHP 8.0.

PR #244, opened 2020-09-17
Update autoloader.inc.php

arshidkv12:


timint:
Thanks, I need to give this thinking. So I am closing this meantime. I would have wanted change the syntax anyway.

arshidkv12:
It needed to load composer library.

timint:
Right, so put composer in ~/vendor/composer/ and the autoloader in ~/vendor/autoload.php and LiteCart should autodetect it.

arshidkv12:
I think, it will make conflict with different modules.

timint:
Should not be a problem. As the Composer autoloader uses prepend for spl_register_autoload().

PR #247, opened 2020-10-13
content CSS classes

litecart-cs:
For better and easier styling by content type.


timint:
This pull request doesn't look right. The typos don't exist in this repo and the class naming of the other files are inconcistent. See STANDARD.md. But I have been thinking of a better way to refer to CSS on specific pages like this: body[data-page="pagename"] .myclass

PR #249, opened 2020-11-05
unification of translation "VAT Registration ID"

litecart-cs:
"VAT Registration ID" or "Tax ID" - Is there a difference?


timint:
They used to be the same but I separated them as too many users renamed "Tax ID" into something else like "organization number" ""civic registration number" etc. And to prevent that in the order copy I separated the two.

PR #253, opened 2020-12-07
Admin enhancements

s22-tech:
Sorted admin module sub-sections alphabetically so they're more easily found;
Added subtitles to the Modules with an explanation of what each module type does.


s22-tech:
More enhancements

PR #277, opened 2021-12-03
Fixed quantity step validator caption

postscripter8:
Fixed quantity step validator caption. When adding to cart an item with a defined quantity step, but the added quantity is not a multiple of that step, an error popup appeared with validator caption saying "You can only purchase sets by %num for this item", but the number was erroneously showing maximum allowed purchase quantity. Now it's the step.


timint:
Big thank you :)

PR #282, opened 2022-02-23
Fix issue where email multipart header is incomplete

mishfit:
Whenever the `[Error Report]` or `[Not Found Report]` emails run I get an additional email from my self hosted mail server (dovecot/postfix):
```
Bad header:
  MIME error: error: part did not end with expected boundary; ; error:
    unexpected end of parts before epilogue
Content type: BadHdrMime
Internal reference code for the message is 1038650-01/2ZhVJIbzk1Da

First upstream SMTP client IP address: [127.0.0.1] 
```

Essentially the multipart boundary declared in the `Content-Type: multipart/mixed; boundary` header is missing at the end of the body's content:
```
Content-Type: multipart/mixed; boundary="==Multipart_Boundary_xxxxxx
```

This change inserts an appropriate ending using the declared multipart boundary if the body is not empty.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,13 @@ ehthumbs.db
 Thumbs.db
 error_log
 .ftpquota
+
+######################
+# VisualStudio Files #
+######################
+.vs/
+
+######################
+# Vim Files          #
+######################
+*.sw[op]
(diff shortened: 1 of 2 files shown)


timint:
Thanks. This does already seem to be resolved in the 2.4.0-dev. The line numbers don't add up.
image

mishfit:
Interesting...I seem to be on a particularly old commit. Yes, this is no longer necessary. Thanks for checking in on that!

timint:
@mishfit Any day near

PR #306, opened 2023-01-02
Product options, inconsistent +/- location

bloodyKnuckles:
There is a discrepancy when adding product options with different prices, and some are more than the base price and some are less than the base price—the  `+` is displayed before the currency symbol sign, and the `-` is displayed after the currency symbol.

This proposal puts both the `+` and `-` before the currency symbol.


timint:
Thank you I will take it :)

PR #307, opened 2023-01-03
Show total product option price when = is selected

bloodyKnuckles:
When setting product options with various prices, and the `=` sign is selected, show the total price of the product having that option rather than the difference between it and the base price. If someone prefers to show the difference they can select `+` or `-`.

### DIFF
--- a/public_html/includes/references/ref_product.inc.php
+++ b/public_html/includes/references/ref_product.inc.php
@@ -322,9 +322,9 @@
 
                   case '=':
                     if ((float)$value[$this->_currency_code] != 0) {
-                      $value['price_adjust'] = currency::convert($value[$this->_currency_code], $this->_currency_code, settings::get('store_currency_code')) - $this->price;
+                      $value['price_adjust'] = currency::convert($value[$this->_currency_code], $this->_currency_code, settings::get('store_currency_code'));
                     } else {
-                      $value['price_adjust'] = $value[settings::get('store_currency_code')] - $this->price;
+                      $value['price_adjust'] = $value[settings::get('store_currency_code')];
                     }
                     break;
 


timint:
Thank you for contributing. I had that idea myself but as that makes it incompatible with multiple option groups I decided to go the other way of displaying the difference. I'm afraid I will have users complain if I roll this out.

bloodyKnuckles:
Rather than altering `price_adjust`, would adding an additional value, such as `price_fixed`, be compatible with multiple option groups?

timint:
@bloodyKnuckles No, as that setting would be per attribute option value. And a product could have multiple attribute option groups.

So it would be odd to pick a fixed price like:
Frame: Oak $9.95
Material: Cotton $9.95

As this would make more sense:
Frame: Oak +$0.95
Material: Cotton +$1.95

bloodyKnuckles:
My use case is like:

T-Shirt

size:
• Small — $15.00
• Medium — $17.00
• Large — $17.00
• X-Large — $20.00

color:
• Red
• Green
• Blue

Am I using the wrong admin control to set this up?

timint:
So your use case is perfectly safe as only one of the groups use price adjustments. My point is multiple groups can't have fixed prices as they conflict eachother. Whick price is it, the fixed price of group A or group B.

The part where the small t-shirt have a negative price operator I would say is an optimization problem. Your base price should probably be more conveniently set to $2 less.

o Small
o Medium +$2
o Large +$2
o X-Large +$5

bloodyKnuckles:
I prefer the shopper not have to do math, as simple as add $2 to the base price may be.

Do know or suspect of a program conflict with adding an additional variable, alongside `price_adjust`, such as `price_subtotal` that provides the total price, for example, a medium size t-shirt?

timint:
> I prefer the shopper not have to do math
What do you mean, the final price of the selected options is displayed in bold 175% bigger right beneath.

image

With fixed prices, isn't the customer forced to do math by "how much more expensive is this option?".

bloodyKnuckles:
> ...the final price of the selected options is displayed in bold 175% bigger right beneath.

It may be nitpicky, but I prefer to present the user with:

T-Shirt

size:
o Small — $15.00
• Medium — $17.00
o Large — $17.00
o X-Large — $20.00

**$17.00**
Excluding Tax

...no brainer.

> With fixed prices, isn't the customer forced to do math by "how much more expensive is this option?".

True, if the user has to choose between options, then I agree use increment/decrement. But I know I where a large t-shirt and don't care how much less a small is or more a x-large it—I just wanna know how much am I going to be paying for a large.

> How do we determine which option value's subtotal by which group to be used?

I'm using the `price_operator` variable having the value `=` assigned to it to display `price_subtotal` next to the product option, instead of `price_adjust`.

> ...other option values conflicting it.

In the case of the T-Shirt example, do you see any potential programming conflicts?

timint:
> In the case of the T-Shirt example, do you see any potential programming conflicts?
No, becuase you are just using one attribute group Sizes. There is no other group to cause a conflict.

I'm closing this as I don't want to roll out something that will cause conflicts. But I think this can be a good add-on for those who want it.

bloodyKnuckles:
> I'm closing...

I understand.

Since I'm not familiar with the code, please clarify:

> ...you are just using one attribute group Sizes.

Are you aware of a potential conflict with two groups: "Sizes" and "Color"?

T-Shirt

size:
o Small — $15.00
o Medium — $17.00
• Large — $17.00
o X-Large — $20.00

color:
o Red
o Green
• Blue

**$17.00**
Excluding Tax

IBBoard:
If anyone wants to implement this as a vmod, I'm using this mod, which seems to work:

```
  <file name="pages/product.inc.php">
    <operation method="before" type="inline" onerror="warning">
      <!-- If we set an "exact price" option then display them as such -->
      <find><![CDATA[if ($value['price_adjust']) {]]></find>
      <insert><![CDATA[            if ($value['price_operator'] == '=') {
              $price_adjust_text = ' '.currency::format(tax::get_price((!empty($product->campaign['price']) ? $product->campaign['price'] : $product->price) >
            }
            else ]]></insert>
    </operation>
  </file>
```

I may release it as an addon at some point. For now, this is just copied from the middle of a larger vmod and needs the appropriate additional tags.

This doesn't address the problem of multiple fixed-price groups, but a) I only have one fixed price group so it Works For Me™ and b) it feels like a broader problem than just display - what *is* the price if someone picks two different fixed price options? It's like saying that you want it both black _and_ white, or size XS _and_ XL. It just isn't a meaningful situation.

PR #313, opened 2023-04-25
add attribute_groups to display them in settings pages

mishfit:
I have a module for which I'd like to disable certain options (`attribute_group`).

I need these lines of code so I can add them to the payment module's settings.


timint:
Very good, thank you.

PR #326, opened 2023-11-11
Checking stock via ajax gives an error when product is out of stock but still orderable

Mik3yZ:
There was no statement for out of stock but orderable.


timint:
Thank you :)

PR #358, opened 2025-03-20
Allow trailing slash on Contact Us URL

IBBoard:
All pages except the Customer Services contact page allowed an optional trailing slash on the URLs. Two chars in the route regex seems to make it consistent.


timint:
Thank you :)

Issue #359, opened 2025-03-29
Litecart doesn't support full UTF-8 in database


IBBoard:
### What where you trying to do?

In any text field that gets saved to the database, enter a value from outside the Base Multilingual Plane. Save the value.

e.g. add a product category called `☺😂` or `Nightware 🌙`

### What actually happened?

LiteCart (actually MySQL/MariaDB) silently mangles the non-BMP text and displays `????` instead.

e.g. `☺????` because `☺` is in the BMP or `Nightware ???`

### What did you expect to happen?

LiteCart uses utf8mb4 to support all of Unicode and the string is stored and retrieved correctly - i.e. `☺😂` or `Nightware 🌙`

MySQL's `utf8_…` collations are flawed and don't support anything outside the BMP. This includes some [supplementary language scripts and emoji](https://en.wikipedia.org/wiki/Plane_(Unicode)#Overview) ([info](https://mathiasbynens.be/notes/mysql-utf8mb4))

### LiteCart Version

LiteCart 2.6.2

### PHP Version

PHP 8.3.6

### Error/Backtrace

```shell

```

### Is this a problem that can be reproduced in the demo platform (https://demo.litecart.net/admin)?

Yes

### If this a visual problem that could be related to a browser. Which one?

None

IBBoard:
I haven't actually recreated this in the demo, because the demo is giving me a blank page right now, but it's a technical limitation of MySQL and MariaDB, so it should be reproducible.

This then seems to have a knock-on effect on a change in 2.6.2 where the new `product_attribute` index on `lc_products_attributes` throws an error because the index can't be more than 1000 bytes. I found this out because I used DOM inspector when I first installed and injected a utf8mb4 encoding, because that's what I always use due to the MySQL limitation.

I'm guessing the new index is because LiteCart wants to allow an existing attribute to be repeated with multiple custom value names? I'm not sure what use case there is for that versus having your product attributes set up so that it contains everything you need in the list.

timint:
Closing as utf8mb4 is the new standard in 3.0 and 2.x will be discontinued.
2.x users can go Admin -> Languages -> Storage Encoding and select utf8mb4

PR #360, opened 2025-03-29
Bug-359: Full UTF-8 support

IBBoard:
Switches LiteCart to use `utf8mb4` character sets and collations, which have existed since MySQL 5.5.

I've included an automatic migration script, because it's safe to go from `utf8mb3` to `utf8mb4` as the latter is a superset of the former. I can rewrite it if that's not the right way to do it. I've tried to apply it where we can, but avoid changing things if it's not a broken-UTF-8 --> good-UTF-8 migration.

As an added bonus, [MySQL 9 is expected to drop utf8mb3 support](https://dev.mysql.com/doc/refman/8.4/en/charset-unicode-utf8.html), so this gets LiteCart ahead of the curve 🙂

Addresses #359

### DIFF
--- a/public_html/includes/library/lib_database.inc.php
+++ b/public_html/includes/library/lib_database.inc.php
@@ -132,7 +132,7 @@
         'iso-2022-jp' => 'cp932',
         'iso-2022-jp-2' => 'eucjpms',
         'iso-2022-kr' => 'euckr',
-        'utf-8' => 'utf8',
+        'utf-8' => 'utf8mb4',
         'utf-16' => 'utf16',
         'windows-1250' => 'cp1250',
         'windows-1251' => 'cp1251',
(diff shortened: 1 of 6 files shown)


timint:
This is already implemented in 3.0. Otherwise I would happily like to accept this commit.
I'm working hard to finalize many things in 3.0 so it can be released. Can I interest you in an early access to 3.0?

IBBoard:
I'd seen you mention 3.0 previously, but didn't know how far along it was. And I hadn't seen a 3.0 branch.

I'm still in the "migrating a site from Wordpress/WooCommerce" phase, and it won't re-open for a couple of months, so I can definitely have an early look at 3.0. Would it be better to base my Content Security Policy stuff on that as well?

timint:
I am working on getting a mostly working version ready. Reach out to me on email and I can send it to you for studying.

PR #377, opened 2025-07-22
Implement module hook system for extensibility

arshidkv12:
Modules can use the `Hooks` class to extend or modify LiteCart functionality without changing core files.

To add a hook in your module:

1. Create a hook file in your module:

`modules/your_module/hooks/my_hook.php`

2. Add your custom logic using `Hooks::add_action()` or `Hooks::add_filter():`

```Hooks::add_action('before_render_head', function() {
  echo '<meta name="my-hook" content="custom">';
});

Hooks::add_filter('product_price_display', function($price) {
  return $price * 0.9; // Apply 10% discount
});
```

3. Ensure hook files are autoloaded by scanning all:
`modules/*/hooks/*.php`

### DIFF
--- a/public_html/includes/app_header.inc.php
+++ b/public_html/includes/app_header.inc.php
@@ -44,3 +44,12 @@
   stats::$data['before_content'] = microtime(true) - SCRIPT_TIMESTAMP_START;
 
   stats::start_watch('content_capture');
+
+  $modules_hooks_dirs = glob(FS_DIR_APP . 'includes/modules/*/hooks', GLOB_ONLYDIR);
+
+  foreach ($modules_hooks_dirs as $hook_dir) {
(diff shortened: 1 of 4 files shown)


timint:
Hi, we already have hook events in LiteCart but no hook module loader as we have vmod. I was considering a hook loader but realized it's just going to be more limited than using vmod directly.

```
event::register('product_price_display', function($price) {
  return $price * 0.9; // Apply 10% discount
})
```

`event::fire('product_price_display', $product['price']);`

Returning a result from a hook is harder as there is no functionality piping the result from one event handler to the next. This is due to argument I/O inconsistencies.

arshidkv12:
Thank you

timint:
Closing this but I appreciate the efforts of trying to come up with new ideas. :)

PR #378, opened 2025-07-23
Update app.less (fix featherlight content background dark mode)

Nacho-AT:
FIX: add !important to featherlight content for dark mode


timint:
Thank you 👍

PR #384, opened 2025-09-17
Update edit_order.inc.php

s22-tech:
Prevent errors when there are no orders yet in the db.


timint:
I'll take it, thank you :)

PR #393, opened 2026-01-20
Update structure.sql

s22-tech:
After doing an update to all my lc_products_xxx tables, I ended up with multiple copies of the products showing in the cart with the exact same information, including the product_id.

There was only a single entry for each item in lc_products, lc_products_info, and lc_lc_products_to_categories.  However, it turned out that there were multiple prices for the same `product_id` in lc_products_prices.

Changing the product_id key to be unique should prevent that problem from happening.


timint:
Thank you, and an upgrade patch will be committed:

```sql
DELETE t1 FROM `lc_products_prices` t1
JOIN `lc_products_prices` t2 ON t1.product_id = t2.product AND t1.id > t2.id;
-- -----
ALTER TABLE `lc_products_prices`
DROP KEY product_id,
ADD UNIQUE KEY product_id (product_id);
```

PR #395, opened 2026-03-17
* Fix #350 - Footer settings and responsive layout

### LINKED ISSUE #350: Footer doesn't work well on cell phones
### What where you trying to do?
Open in a cellphone

### What actually happened?
shows only 2 columns in the footer

### What did you expect to happen?
it should show 5 columns, 2 on each row

### LiteCart Version

2.6.1

### PHP Version

PHP 8.3

### Error/Backtrace

_No response_

### Is this a problem that can be reproduced in the demo platform (https://demo.litecart.net/admin)?

Yes

### If this a visual problem that could be related to a browser. Which one?

None

Wachhund:
## Summary

- Add configurable footer section visibility (Categories, Manufacturers, Contact) via **template settings**
- Add configurable limits for Categories and Manufacturers entries (default: 9, with "More..." link)
- Replace CSS breakpoint hiding (`hidden-xs`, `hidden-sm`) with settings-based toggles
- Improve responsive footer grid: 2 columns mobile, 3 sm, 5 md+
- Add text-overflow ellipsis for long footer entries

## Context

Addresses #350 — "Footer doesn't work well on cell phones". As suggested by @timint, `hidden-xs` classes are removed. Footer section visibility is now configurable via template settings (`config.inc.php`), accessible under **Admin > Appearance > Template Settings**.

## Template Settings Added

| Key | Default | Description |
|-----|---------|-------------|
| `footer_categories` | On | Toggle categories section |
| `footer_categories_limit` | 9 | Max categories shown |
| `footer_manufacturers` | On | Toggle manufacturers section |
| `footer_manufacturers_limit` | 9 | Max manufacturers shown |
| `footer_contact` | On | Toggle contact section |

Settings are template-specific — different templates can have their own footer configuration independently. No global settings registry entries or upgrade patches needed.

## Test Plan

- [ ] Footer settings appear in Admin > Appearance > Template Settings
- [ ] Toggling sections on/off hides/shows them in the storefront footer
- [ ] Limits control the number of entries, "More..." link appears when exceeded
- [ ] Footer is responsive: 2 columns on mobile, 3 on tablet, 5 on desktop
- [ ] Long entry names truncate with ellipsis
- [ ] All defaults match current behavior (no visual change on existing installations)


timint:
Hi, happy to see some work being made and submitted. 👍
I am thinking of this should perhaps be moved to the template settings rather than settings registry. As it is pretty much template specific.

Wachhund:
Good point! I've moved the footer settings from the global settings registry to the **template settings** (`config.inc.php`). They now appear under **Admin > Appearance > Template Settings** alongside the existing sidebar/category tree options.

Changes in the updated commit:
- Footer settings defined in `templates/default.catalog/config.inc.php`
- Box/view code reads from `document::$settings` instead of `settings::get()`
- Removed the `footer` settings group from `data.sql` and the `2.6.5.sql` upgrade patch (no longer needed — template settings auto-initialize from defaults)
- Default values included in `store_template_catalog_settings` JSON for new installations

This way, different templates can have their own footer configuration independently.

timint:
I will take it :)

PR #396, opened 2026-03-19
* Fix header warning when accessing install.php after installation

Wachhund:
## Summary

Fixes a PHP warning when accessing `install.php` after the installation is complete:

```
Warning: Cannot modify header information - headers already sent
(output started at includes/header.inc.php:1)
in install/install.php on line 46
```

## Cause

When `install.php` is accessed without the `install` POST parameter, it redirects to `index.php` via `header('Location: ...')`. However, the HTML header template (`includes/header.inc.php`) is included *before* the redirect check, so output has already been sent when `header()` is called.

## Fix

Move the redirect check (`if (empty($_REQUEST['install']))`) before the `require header.inc.php` include, so the `Location` header is sent before any HTML output.

## Test Plan

- [ ] Access `install.php` directly after installation — should redirect cleanly to `index.php` without warnings
- [ ] Run the installer normally — installation process works unchanged
- [ ] CLI installation (`php install.php --install ...`) still works

### DIFF
--- a/public_html/install/install.php
+++ b/public_html/install/install.php
@@ -39,12 +39,13 @@
     }
 
   } else {
-    require __DIR__ . '/includes/header.inc.php';
-  }
 
-  if (empty($_REQUEST['install'])) {
-    header('Location: index.php');
-    exit;
+    if (empty($_REQUEST['install'])) {
+      header('Location: index.php');
+      exit;
+    }
+
+    require __DIR__ . '/includes/header.inc.php';
   }
 
   ob_start();


timint:
Thank you for being awesome! 👍

Wachhund:
Thanks! Happy to contribute. More to come 🐾

timint:
You will also find LiteCart 3.0 in dev-major branch.
Have a peek. :)

Wachhund:
> "Have a peek. :)"

Peeked. Now I can't un-peek. You should put a warning label on that branch. 🔥

timint:
😄
In v3 you will find that cart, checkout, and products are partially broken. This is due to "stock items" where I have not (yet) had my final thoughts.

Wachhund:
Took a deeper look at the stock items system. Opened #398 with the findings — 5 concrete bugs causing the cart/checkout/products breakage. Working on a fix branch.

PR #399, opened 2026-03-19
Fix 5 stock items integration bugs (closes #398)

### LINKED ISSUE #398: Stock items integration: 5 bugs causing cart/checkout/products breakage (dev-major)
Following up on the [dev-major peek](https://github.com/litecart/litecart/pull/396#issuecomment-4087269758) — here are the root causes of the cart/checkout/products breakage mentioned by @timint.

## 5 concrete bugs in the stock items integration

| # | File | Bug |
|---|---|---|
| 1 | `ent_stock_item.inc.php:223` | References non-existent table `products_to_stock_items` — should be `products_stock_options` |
| 2 | `nod_cart.inc.php:241` | Stores `stock_option.id` (PSO primary key) instead of `stock_option.stock_item_id` — wrong reference propagates to orders |
| 3 | `ent_order.inc.php:317-319` | Stock restitution updates `lc_products` instead of `lc_stock_items` |
| 4 | `ent_order.inc.php:293` | `notes` field gets populated with `ip_address` value |
| 5 | `shopping_cart.inc.php:186` | `var_dump()` left in production code |

## Incomplete areas

- `ent_product` has load logic for stock_options but no save logic
- Bundle product type (`stock_option_type = 'bundle'`) — assignment mechanism unclear
- `price_adjustment` is FLOAT in DB but frontend suggests multi-currency
- Stock reservation on order creation doesn't go through `stock_transactions`

## Next steps

I'll take a shot at fixing the 5 bugs on a separate branch. No promises on the architectural decisions (bundles, price adjustments) — those are yours to make. 🐾

Wachhund:
## Summary

Five bugs in the stock items integration that cause cart, checkout, and product breakage on `dev-major`:

- **Non-existent table reference** (`ent_stock_item.inc.php:223`): `products_to_stock_items` → `products_stock_options`
- **Wrong ID in cart** (`nod_cart.inc.php:241`): Stored `stock_option.id` (PSO primary key) instead of `stock_option.stock_item_id` — wrong reference propagated to orders
- **Notes/IP swap** (`ent_order.inc.php:293`): `notes` field was populated with `ip_address` value
- **Orphaned restock query** (`ent_order.inc.php:315-320`): Query outside the foreach loop targeted wrong table (`products` instead of `stock_items`) with wrong column (`product_id` instead of `id`). Cross-checked with `dev` branch — [v2 only has the foreach loop](https://github.com/litecart/litecart/blob/dev/public_html/includes/entities/ent_order.inc.php#L291-L297), no extra query. This was introduced during the v3 refactor.
- **Debug var_dump** (`shopping_cart.inc.php:186`): Left in production code

## Test plan

- [ ] Add product with stock options to cart — verify correct `stock_item_id` is stored
- [ ] Complete checkout — verify order notes contain actual notes, not IP address
- [ ] Change order status with `stock_action = 'commit'` — verify stock is restocked correctly
- [ ] Verify no `var_dump` output on shopping cart page

### DIFF
--- a/public_html/frontend/pages/shopping_cart.inc.php
+++ b/public_html/frontend/pages/shopping_cart.inc.php
@@ -183,7 +183,6 @@
 
 		if ($cheapest_shipping) {
 			$_page->snippets['cheapest_shipping'] = $cheapest_shipping;
-			var_dump($_page->snippets['cheapest_shipping']);
 		}
 	}
 
--- a/public_html/includes/entities/ent_order.inc.php
+++ b/public_html/includes/entities/ent_order.inc.php
@@ -290,7 +290,7 @@
 					discount_tax = ". (float)$this->data['discount_tax'] .",
 					total = ". (float)$this->data['total'] .",
 					total_tax = ". (float)$this->data['total_tax'] .",
-					notes = '". database::input($this->data['ip_address']) ."',
+					notes = '". database::input($this->data['notes']) ."',
 					conversions = '". database::input(f::format_json($this->data['conversions'])) ."',
 					ip_address = '". database::input($this->data['ip_address']) ."',
 					hostname = '". database::input($this->data['hostname']) ."',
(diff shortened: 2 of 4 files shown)


timint:
Very good. 3.0 moving forward!👍🙂
Let me know if you find the new 3.0 framework and simplifications welcoming.

(Off topic: You probably figured out already that the new nod_session handler does not support serializing the ent_order object. Which is one of the reasons the checkout becomes broken. I don't think we need to serialize the entire object. just $order->data.)

Wachhund:
Thanks for the hint about `nod_session`! That explains a lot — I saw the checkout breaking but traced it to the stock items integration first.

Serializing only `$order->data` makes sense — the full entity carries database connections and closures that don't survive serialization. Should this be handled in `ent_order` (e.g. `__serialize()` / `__unserialize()`) or on the session side?

Would be happy to look into this as a next step.

Wachhund:
Following up on the serialization question — would you prefer handling it in `ent_order` via `__serialize()`/`__unserialize()`, or on the session side in `nod_session`? Happy to take either approach.

PR #401, opened 2026-03-19
Add GitHub Actions CI for dev-major

Wachhund:
## Summary

Three-stage CI pipeline for `dev-major`:

**1. Lint** (blocking)
- PHP syntax check on all files
- JSON validation
- SQL file validation

**2. Build** (blocking)
- Gulp asset compilation (LESS, SCSS, JS) with Node.js 24
- Validates that all stylesheets and scripts compile correctly (also covers LESS validation with full import chain — a standalone LESS lint was removed because it can't resolve shared variables in isolation)

**3. Platform Tests** (non-blocking)
- MariaDB 11 service container
- Config generated from install template, schema imported from `structure.json` + `data.sql`
- Runs all 30 entity CRUD tests from `tests/`
- Marked `continue-on-error` because v3 is still in development (e.g. `brands` table missing `image` column)

**Additional scanners** (non-blocking, report only):
- **Debug statement scanner**: Finds `var_dump`/`print_r` left in production code — currently reports 3 instances
- **Insecure PRNG scanner**: Flags `mt_rand`/`str_shuffle` usage — currently reports 10 instances

The goal was to get CI infrastructure in place, not to fix the issues it finds. The non-blocking checks show known issues as warnings — once resolved, they can be made blocking.

Also fixes a trailing comma in `install/data/default/public_html/vendor/composer.json` (invalid JSON).

## Test plan

- [x] PHP Lint: all files pass
- [x] JSON Lint: all files pass (after composer.json fix)
- [x] SQL Lint: all files pass
- [x] Gulp Build: all assets compile
- [x] Platform Tests: run through (3/30 pass before known schema issue)
- [x] Debug scanner: correctly identifies 3 var_dump statements
- [x] PRNG scanner: correctly identifies 10 insecure calls

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,7 +68,7 @@ jobs:
           node-version: '24'
 
       - name: Install dependencies
-        run: npm ci
+        run: npm install --no-audit --no-fund
 
       - name: Build assets
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,7 +119,9 @@ jobs:
       - name: Install LiteCart via CLI installer
         working-directory: public_html/install
         run: |
-          php install.php \
+          php -d "auto_prepend_file=" \
+            -r "\$_SERVER['SERVER_SOFTWARE']='CLI'; \$_SERVER['REQUEST_METHOD']='GET'; require 'install.php';" \
+            -- \
             --document_root="$(realpath ../../public_html)" \
             --db_server="127.0.0.1" \
             --db_username="litecart" \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,9 @@ name: CI
 
 on:
   pull_request:
-    branches: [dev, dev-major]
+    branches: [dev-major]
   push:
-    branches: [dev, dev-major]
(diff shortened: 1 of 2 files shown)


timint:
I see you replaced that with direct/manual config instead. What are your thoughts around this? Using CLI for installing didn't sound like a bad idea.

Wachhund:
The CLI installer is actually the better approach — I'd prefer using it. The reason I switched to direct config was a chicken-and-egg issue in `install.php`:

Line 10 checks `$_REQUEST['document_root']`, but `$_REQUEST` is only populated by `getopt()` at line 72. In CLI mode, `$_REQUEST` is empty until then, so the DOCUMENT_ROOT detection always fails. Also, `$_SERVER['SERVER_SOFTWARE']` isn't set in PHP CLI by default — the check on line 39 needs it to be exactly `'CLI'`.

Happy to submit a fix for the installer if you'd like — then CI can use it properly.

timint:
Sure, I welcome a fix. You are very helpful. CLI detection in PHP is annoying as php_sapi_name() is pretty useless to rely on. It's compatibility.inc.php line 27 that sets this value for $_SERVER['SERVER_SOFTWARE']. Not very industry standard but it was my best shot I had.

PR #407, opened 2026-03-22
* Harden MCP server security and add search tools

Wachhund:
The OWASP Secure MCP Server Development Guide made for some bedtime reading. Here's what fell out.

## Security fixes

| Change | Impact |
|--------|--------|
| Authentication mandatory | Previously, requests without credentials silently bypassed auth. Now returns 401. |
| Request body limit (64KB) | Prevents memory exhaustion from oversized payloads |
| Dead code removed | Unreachable `throw` after if/else block in both endpoints |
| Tool discovery separated by role | Backend searches `backend/mcp/`, frontend searches `frontend/mcp/` — customers can no longer discover admin tools |

## New tools

| Tool | Scope | Description |
|------|-------|-------------|
| `products_search` | Admin | Search by keyword, category, brand, status (max 50) |
| `orders_list` | Admin | Filter by status, customer, date range (max 50) |

Both read-only, both require authentication.

## Reference

Security changes based on [OWASP Secure MCP Server Development](https://genai.owasp.org) — specifically sections on architecture, data validation, and authentication.

## Test plan

- [ ] Unauthenticated request returns 401 (both endpoints)
- [ ] Authenticated `tools/list` on backend shows `ping`, `stats_summary`, `products_search`, `orders_list`
- [ ] Authenticated `tools/list` on frontend shows only `ping`
- [ ] `products_search` returns products filtered by keyword
- [ ] `orders_list` returns orders filtered by date range
- [ ] Request body > 64KB is truncated/rejected
- [ ] PHP lint passes

### DIFF
--- a/public_html/backend/mcp/mcp_orders_list.inc.php
+++ b/public_html/backend/mcp/mcp_orders_list.inc.php
@@ -1,0 +1,92 @@
+<?php
+
+	return $schema = [
+		'name' => 'orders_list',
+		'description' => 'List recent orders with status, customer, total, and date. Supports filtering by status, customer, and date range.',
+		'inputSchema' => [
+			'type' => 'object',
+			'properties' => [
+				'order_status_id' => [
+					'type' => 'integer',
+					'description' => 'Filter by order status ID',
+				],
+				'customer_id' => [
+					'type' => 'integer',
+					'description' => 'Filter by customer ID',
+				],
+				'date_from' => [
+					'type' => 'string',
+					'description' => 'Filter orders created on or after this date (YYYY-MM-DD)',
+				],
+				'date_to' => [
+					'type' => 'string',
+					'description' => 'Filter orders created on or before this date (YYYY-MM-DD)',
+				],
+				'limit' => [
+					'type' => 'integer',
+					'description' => 'Max results (default: 10, max: 50)',
+				],
+			],
+			'required' => [],
+		],
+	];
+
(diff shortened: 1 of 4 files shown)


timint:
The frontend mcp server was at first thought to be public, without auth. Then I had a second thought that it might not be a good idea. Then I had a third idea maybe supporting both was a good idea depending on the tool offered. I also started sketching on a UCP implementation but stopped as ACP, MCP, and UCP all transition very fast.

(Sidenote: The most recent MCP standard uses oAuth 2.x. For simplicity and minimalistic thinking I decided to go with the protocol of 2024-11-05. But I hear there should be some support for Bearer tokens in the latest MCP protocol. Still unsure if it should be customers, merchants or custom mcp users who use this endpoint.)

Wachhund:
Good thoughts — I've worked with both Bearer tokens and OAuth 2.x before, and it really comes down to the target audience. For shop owners and developers integrating their own store, Bearer tokens are simple and effective — generate a token in admin, add it to your MCP client config, done.

I'd suggest starting with Bearer token auth as a first step. OAuth can be layered on top later if needed — for example by appending an `/oauth` path segment to the MCP endpoint that triggers the OAuth flow, while the base endpoint keeps working with Bearer tokens. That way you don't over-engineer it upfront but have a clean upgrade path.

As for who should use it — merchants (admin API) and custom MCP users seem like the primary audience. Customer-facing tools could work too, but that's a different trust level and scope.

PR #408, opened 2026-03-22
* Fix checkout session serialization for ent_order

Wachhund:
Following up on the conversation in #399 — this should fix the checkout breakage caused by `ent_order` not surviving session serialization.

## Problem

`session::save()` serializes session data via `json_encode()`. The checkout stores a full `ent_order` object in `session::$data['checkout']['order']`. This object carries `$shipping` and `$payment` properties containing module instances with closures and DB connections — none of which survive `json_encode()`.

## Fix

| File | Change |
|------|--------|
| `ent_order.inc.php` | `implements JsonSerializable` — `jsonSerialize()` returns only `$data`, skipping `$shipping`, `$payment`, `$previous` |
| `ent_order.inc.php` | `from_data(array $data)` — static factory to rebuild an `ent_order` from a plain array |
| `nod_session.inc.php` | `load()` checks if `checkout.order` is an array (from JSON) and restores it via `ent_order::from_data()` |

The cycle: `ent_order` → `jsonSerialize()` → only `$data` → JSON in DB → `json_decode()` → array → `from_data()` → `ent_order` with `$data` restored. Shipping and payment modules are re-attached during checkout processing, not during session restore.

## Test plan

- [ ] Add products to cart, proceed to checkout — order persists across page loads
- [ ] Close browser, reopen — checkout session is restored
- [ ] Complete checkout flow (customer details → shipping → payment → confirm)
- [ ] `json_encode(session::$data)` no longer throws on order objects
- [ ] PHP lint passes

### DIFF
--- a/public_html/includes/entities/ent_order.inc.php
+++ b/public_html/includes/entities/ent_order.inc.php
@@ -1,6 +1,6 @@
 <?php
 
-	class ent_order {
+	class ent_order implements JsonSerializable {
 		public $data;
 		public $previous;
 
@@ -1022,5 +1022,17 @@
 			cache::clear_cache('category');
 			cache::clear_cache('brand');
 			cache::clear_cache('products');
+		}
+
+	// Serialize only $data for session storage (json_encode compatibility)
+		public function jsonSerialize(): mixed {
+			return $this->data;
+		}
+
+	// Restore from session data
+		public static function from_data(array $data): self {
+			$order = new self();
+			$order->data = $data;
+			return $order;
 		}
 	}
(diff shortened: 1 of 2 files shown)


timint:
That's not a bad way to do it. I thought of JsonSerialized myself but have memory of it not being capable of restoring object methods. The way json is being used might be obsolete as we have json serialization already in the custom session handler. But I like the static functions you added. Let's play with this and see how we like it. 👍

Wachhund:
Good point about the JSON serialization in the session handler — if `nod_session` already handles JSON, the `__serialize()`/`__unserialize()` on the entity might be redundant. Happy to let it ride as-is and see how it plays out.

PR #410, opened 2026-03-22
* Harden HTTP security headers, add per-user error display toggle

Wachhund:
Ran the response headers through a security scanner. It politely suggested we stop introducing ourselves to every bot on the internet.

## What's changed

| Change | Why |
|--------|-----|
| Add `X-Content-Type-Options: nosniff` | Prevents MIME type sniffing attacks |
| Add `Referrer-Policy: strict-origin-when-cross-origin` | Limits referrer leakage to external sites |
| Add `Strict-Transport-Security` (HTTPS only) | Tells browsers to always use HTTPS once seen |
| Remove `X-Powered-By` header | No need to advertise the platform name to scanners |
| Remove session ID from `_env` JS object | Session IDs don't belong in client-side JavaScript |
| Add `session.cookie_secure` flag | Prevents session cookies from being sent over plain HTTP |
| Per-user `show_php_errors` toggle | Replaces the blanket `display_errors = On` for all admins — now opt-in per user account (off by default) |

## Per-user error display (updated per your feedback)

Instead of removing `display_errors` entirely, admin users now get a `show_php_errors` column (default: off). When enabled, PHP errors are shown for that user — same visibility as before, but opt-in instead of forced. The error handler continues to use `display_errors` ini as the trigger.

The column is added to `lc_users` in `structure.sql`. The admin UI for the toggle (checkbox in user settings) can be added as a follow-up.

## Test plan

- [ ] Check response headers contain the new security headers
- [ ] Verify `X-Powered-By` is absent from responses
- [ ] Confirm `_env.session.id` is no longer in page source
- [ ] Confirm session cookie has `Secure` flag on HTTPS
- [ ] Admin with `show_php_errors = 0`: no PHP errors visible
- [ ] Admin with `show_php_errors = 1`: PHP errors visible as before


timint:
🙂 I support this.

A give or take is the display_errors=Off for admins. It has been useful for site owners to understand when there is an issue on the page and it has been appreciated to show them. But I also understand the reason why it makes sense to disable it. How about a setting in the admin user account to show or hide php errors when logged in? That would make it up to the end-user to decide what they prefer or need.

Wachhund:
That's a great middle ground. A per-user setting like `show_php_errors` in the admin account would give site owners the diagnostic visibility they're used to, while keeping it off by default for security. I'll update the PR to add that toggle instead of the hard removal.

PR #411, opened 2026-03-22
+ Add CSRF protection for all state-changing requests

Wachhund:
Every form was just... trusting. Like a shop with no lock on the door. Time to hand out keys.

## How it works

| Layer | What happens |
|-------|-------------|
| **Session** | `session::csrf_token()` lazily generates a 32-byte hex token per session |
| **Forms** | `form_draw_form_begin()` auto-injects a hidden `csrf_token` field for POST forms |
| **AJAX** | jQuery `ajaxPrefilter` sends `X-CSRF-Token` header on non-GET requests |
| **Guard** | `app_header.inc.php` validates the token on POST/PUT/DELETE/PATCH before any route runs |
| **Login/Logout** | Token rotation via `session::rotate_csrf_token()` on all 4 login/logout handlers |

## Design decisions

- **Excluded path**: `order_process` is skipped — payment gateway callbacks (PayPal IPN, Stripe webhooks, etc.) don't carry session tokens
- **Constant-time comparison**: Uses `hash_equals()` to prevent timing attacks
- **Backwards compatible**: Existing forms using `form_draw_form_begin()` get CSRF protection automatically — no template changes needed
- **Graceful error**: AJAX gets JSON error, regular requests get a human-readable message with back link

## Test plan

- [ ] Submit any admin form (e.g. edit a product) — should work normally
- [ ] Submit a storefront form (e.g. add to cart, login) — should work normally
- [ ] Remove the csrf_token hidden field from a form via devtools, submit — should get 403
- [ ] Verify AJAX requests include the `X-CSRF-Token` header (devtools network tab)
- [ ] Login/logout and verify the token changes (check `_env.csrf_token` in page source)
- [ ] Test a payment callback flow (e.g. PayPal IPN) is not blocked


timint:
Thank you, we had this years ago and it was removed in maybe the LiteCart 1.x era. That was due to payment providers and some third parties wasn't able to send form posts with it enabled. It's 2026 and I feel CSRF should be added back in. Previously the url routing modules provided a setting post_security => false or a check for a definition e.g. DISABLE_CRSF_SECURITY_TOKEN=1, where it should be disabled. I haven't gotten around to what the best practice is.

Wachhund:
Good to know about the 1.x history! The `post_security => false` approach per route module makes a lot of sense — it's basically what the current `$csrf_excluded_paths` array does, but configurable per module instead of hardcoded. I could refactor it so that route modules (or payment modules specifically) can declare themselves as CSRF-exempt via a property or setting. That way payment callbacks from Stripe, PayPal etc. work out of the box without needing to know the URL pattern upfront.

timint:
I'm gonna accept this pull request as-is because it speaks my language how I would come up with a good base. And we can discuss some ideas for webhooks and callbacks to bypass it. There is a bit of hen and the egg issue as we wanna validate the requests as early in code as possible. But before allow code to flag an incoming request for bypass.

Wachhund:
Makes sense — accept it now, iterate on the bypass mechanism. One idea for the chicken-and-egg: early in bootstrap (before the CSRF check), scan registered route modules for a `csrf_exempt` flag or method. Something like `route::$current->csrf_exempt` that modules can set during registration. That way the module declares the exemption, not the framework hardcoding paths.

PR #412, opened 2026-03-22
* Replace SHA1 remember-me with HMAC-SHA256 tokens

Wachhund:
SHA1 walked into a bar. The bouncer said "I know you, you're weak." So we replaced him with HMAC-SHA256 — same door, better bouncer.

## What's changed

The remember-me cookie for both admin users and customers now uses HMAC-SHA256 instead of plain SHA1 checksums.

| Before | After |
|--------|-------|
| `username:sha1(username + password_hash + IP + UA)` | Base64-encoded JSON with HMAC-SHA256 signature |
| Breaks when IP or User-Agent changes | Stable across networks and browser updates |
| No expiration in token | Built-in expiry timestamp, verified server-side |
| Uses `!=` comparison | Uses `hash_equals()` (constant-time) |
| Works without config | Requires `DB_SERVER_SECRET` in config (graceful skip if missing) |

## How it works

1. On login with "remember me", `token_create_remember()` builds a payload: `{id}:{expiry}:{password_hash_prefix}` and signs it with `hash_hmac('sha256', payload, DB_SERVER_SECRET)`
2. Cookie contains base64-encoded JSON: `{"id": 1, "exp": 1745337600, "sig": "abc..."}`
3. On next visit, `token_verify_remember()` decodes the cookie, checks expiry, rebuilds the expected signature from the DB password hash, and compares with `hash_equals()`
4. Old-format cookies (`email:sha1...`) fail JSON decode and are silently cleared — no user-facing errors during migration

## Additional improvements

- Customer remember-me now checks `status` and `date_blocked_until` before auto-login
- Remember-me cookies get `Secure` flag on HTTPS connections
- Expiry configurable via `remember_me_days` setting (default: 30 days, was hardcoded 3 months)

## Test plan

- [ ] Login with "remember me" checked — verify cookie is set (base64 JSON, not `email:sha1`)
- [ ] Close browser, reopen — auto-login should work
- [ ] Change password — old remember-me cookie should fail (password hash changed)
- [ ] Wait for token to expire (or manually set short expiry) — should not auto-login
- [ ] Test with `DB_SERVER_SECRET` undefined — remember-me should be silently disabled
- [ ] Test with old-format cookie still in browser — should clear it and show login page
- [ ] Admin and customer login/remember-me both covered

### DIFF
--- a/public_html/admin/login.php
+++ b/public_html/admin/login.php
@@ -172,9 +172,10 @@
 
       user::load($user['id']);
 
-      if (!empty($_POST['remember_me'])) {
-        $checksum = sha1($user['username'] . $user['password_hash'] . $_SERVER['REMOTE_ADDR'] . ($_SERVER['HTTP_USER_AGENT'] ? $_SERVER['HTTP_USER_AGENT'] : ''));
-        header('Set-Cookie: remember_me='. $user['username'] .':'. $checksum .'; Path='. WS_DIR_APP .'; Expires='. gmdate('r', strtotime('+3 months')) .'; HttpOnly; SameSite=Lax', false);
+      if (!empty($_POST['remember_me']) && defined('DB_SERVER_SECRET')) {
+        $expiry_days = (int)(settings::get('remember_me_days') ?: 30);
+        $token = functions::token_create_remember($user['id'], $user['password_hash'], $expiry_days);
+        header('Set-Cookie: remember_me='. $token .'; Path='. WS_DIR_APP .'; Expires='. gmdate('r', strtotime('+'. $expiry_days .' days')) .'; HttpOnly; SameSite=Lax' . (!empty($_SERVER['HTTPS']) ? '; Secure' : ''), false);
       } else if (!empty($_COOKIE['remember_me'])) {
         header('Set-Cookie: remember_me=; Path='. WS_DIR_APP .'; Max-Age=-1; HttpOnly; SameSite=Lax', false);
       }
--- a/public_html/includes/functions/func_token.inc.php
+++ b/public_html/includes/functions/func_token.inc.php
@@ -1,0 +1,38 @@
+<?php
+
+  function token_create_remember($id, $password_hash, $expiry_days=30) {
+
+    if (!defined('DB_SERVER_SECRET')) return '';
+
+    $exp = time() + (int)$expiry_days * 86400;
+    $payload = $id . ':' . $exp . ':' . substr($password_hash, 0, 16);
+    $sig = hash_hmac('sha256', $payload, DB_SERVER_SECRET);
+
+    return base64_encode(json_encode([
+      'id' => (int)$id,
+      'exp' => $exp,
+      'sig' => $sig,
+    ]));
+  }
+
+  function token_verify_remember($cookie_value, $password_hash) {
+
+    if (!defined('DB_SERVER_SECRET')) return false;
+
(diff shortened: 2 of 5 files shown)


timint:
Walked into a bar.. haha😄
Again, I like it that you have this pair of fresh eyes and see things that need attention.

I'm thinking about DB_SERVER_SECRET. It will need a migration patch and be generated and added to the install/config template. The naming DB_SERVER_.. feels a bit off as it's being used for cookes and session. Is this something you have in mind to be shared with and reused for database operations?

Wachhund:
Good point on the naming. `DB_SERVER_SECRET` felt off to me too — it's not really a database secret, it's an application-level signing key used for HMAC tokens right now.

I'd suggest renaming it to something like `APP_SECRET` or `SECRET_KEY` — a general-purpose application secret that can be used for cookies, HMAC signatures, CSRF salts, and anything else that needs a shared key. This is the pattern most frameworks use (Django's `SECRET_KEY`, Laravel's `APP_KEY`, Symfony's `APP_SECRET`).

For the migration: the installer could generate it automatically via `bin2hex(random_bytes(32))` and write it to the config template. Existing installations without the key would just have remember-me silently disabled (which is already how the code handles it — the `defined('DB_SERVER_SECRET')` guard clause skips token creation/verification).

Happy to submit a follow-up PR for the rename + installer integration if you'd like.

timint:
Words that come to mind are ..SIGNING_KEY.. or ..HMAC_KEY... And maybe allow them to be distinguishable if they need roation.

```
define('REMEMBER_ME_SIGNING_KEY', '...');
define('CSRF_SIGNING_KEY', '...');
```
or
```
define('REMEMBER_ME_HMAC_KEY', '...');
define('CSRF_HMAC_KEY', '...');
```
or
```
define('HMAC_KEY_REMEMBER_ME', '...');
define('HMAC_KEY_CSRF', '...');
```

Off-topic: But something that is related is the storing of secrets in raw text. I have had the thought if it's a good idea to create some central encrypted vault for storing secrets and credentials. A set of API credentials, salt, encryption keys and what not? Doing so in a database might be convenient, but poses a new risk of leaks in an XSS attack. With that said I just wanted to plant some thoughts. I have clients who adds reusable third-party API_KEYS to config.inc.php as definitions which is handy but also exposed to the entire platform.

timint:
In regards to the storing of secrets. How about a dedicated system node nod_vault.

This would allow for code to set or get sensitive data;
```
vault::set('keyname', $credentials, 'unlockcode'); // set
$credentials = vault::get('keyname', 'unlockcode'); // get
```

But going with an inline unlockcode might make it difficult or problematic to provide a ui vault/credential manager.

Wachhund:
I like the per-purpose key approach — `HMAC_KEY_REMEMBER_ME`, `HMAC_KEY_CSRF` etc. makes rotation safe (changing one doesn't break the other) and the naming is self-documenting. The `HMAC_KEY_*` prefix reads cleanest to me.

The vault idea is interesting. A `nod_vault` that encrypts at rest and requires an unlock code would solve the "API keys in config.inc.php" problem nicely. The tricky part is key management — where does the vault's own encryption key live? Some options:

- Environment variable (not in config.inc.php, not in DB)
- Derived from a master password on first admin login (never stored)
- Hardware-backed if available (but kills shared hosting compatibility)

Worth exploring as a separate feature. For now, per-purpose HMAC keys in config.inc.php is a solid first step.

PR #415, opened 2026-03-22
+ Add business logic platform tests

Wachhund:
30 entity CRUD tests but no one tested if you can actually buy anything. Fixed that.

## What's new

4 test files covering the core business logic — cart, checkout, stock, and product catalog. All follow the existing platform test pattern (transaction rollback, auto-increment restore, return true/false).

| Test file | What it covers | Tests |
|-----------|---------------|-------|
| `tests/product_catalog.php` | Multilingual JSON fields, category filtering, tax calculation, inactive product filtering | 4 |
| `tests/stock_inventory.php` | Stock options ↔ stock items, deposit/withdrawal transactions, out-of-stock detection, cascade deletion | 5 |
| `tests/cart.php` | Add simple/variant products, update quantity, remove items, stock limit enforcement | 5 |
| `tests/checkout.php` | Order creation, variant order lines, stock tracking, multi-line total calculation, order deletion cleanup | 5 |

## Design decisions

- **Same framework**: No PHPUnit, no new dependencies — just the existing return true/false pattern
- **Self-contained fixtures**: Each test creates its own products, stock items, and orders — no dependency on seed data
- **Cart tests use force mode**: `cart::add_product(..., true)` and `cart::remove(..., true)` to avoid `reload()`/`exit` calls in CLI
- **Checkout without payment module**: Tests create orders directly via `ent_order` + `add_line()` + `save()` — tests business logic, not HTTP flow
- **Transaction isolation**: All tests run inside a transaction and rollback on completion

## Test plan

- [ ] All 4 new tests pass in CI (`platform-tests` job)
- [ ] Existing 30 tests still pass (no regressions)
- [ ] Tests can run independently in any order


Wachhund:
All 34 tests green (30 existing + 4 new). CI passed on all three jobs.

While writing the tests, I found and fixed two pre-existing bugs:

1. **`nod_customer.inc.php:262`** — `$countries[0]['iso_code_2']` should be `$countries[0]`, since `fetch_all('iso_code_2')` returns a flat array of strings, not rows. Only triggers when no country is set via cookie/setting (e.g. CLI context).

2. **`structure.json`** — Products table still had column `image` instead of `default_image`. The 3.0.0 migration renames it, but `structure.json` wasn't updated to match. This caused fresh installations from `structure.json` to miss the renamed column.

Both fixes are included in this PR.

Wachhund:
**Update:** The new tests are doing their job — they're finding pre-existing bugs that were hidden because no one tested these code paths on a fresh database before.

Bugs found and fixed so far (6 total):

| # | File | Bug | Fix |
|---|------|-----|-----|
| 1 | `nod_customer.inc.php:262` | `$countries[0]['iso_code_2']` on flat array | `$countries[0]` |
| 2 | `structure.json` (products) | Column `image` not renamed to `default_image` | Match migration |
| 3 | `ent_product.inc.php:100` | JOIN on `pa.value` (column is `value_id`) | `pa.value_id` |
| 4 | `ent_product.inc.php:510-512` | Missing isset() for optional stock option fields | Guard with isset() |
| 5 | `tests/category.php:68` | `parent_id => 1` self-references on empty table | `parent_id => 0` |
| 6 | `structure.json` (orders_lines) | Missing `serial_number`, `discount_tax`, `weight`, `weight_unit` | Add columns from 3.0.0 migration |

All of these exist in the current codebase — our tests just happen to be the first to exercise these paths against a fresh `structure.json`-based database. Working through them one by one as CI reveals them.

timint:
Soooo much useful stuff here, wohoo!

Wachhund:
Thanks! The tests are already paying for themselves — 7 pre-existing bugs found so far. Follow-up PR #416 fixes another one in the order entity.

PR #419, opened 2026-03-23
* Fix 4 security and correctness bugs

Wachhund:
Found these while taking a deep dive into the v3 code. Four unrelated bugs, all small fixes.

## Fixes

| Fix | File | Bug | Impact |
|-----|------|-----|--------|
| MCP blocked_until | `frontend/pages/mcp.inc.php` | Missing `blocked_until` check in login query | Brute-force account lock bypassable via MCP API |
| MCP blocked_until | `backend/pages/mcp.inc.php` | Same as above for admin MCP endpoint | Same — admin accounts affected too |
| SQL disclosure | `includes/nodes/nod_database.inc.php` | SQL query appended to exception message | Full query text (table names, column names, user input) visible when errors are displayed |
| ArrayAccess | `includes/abstracts/abs_reference_entity.inc.php` | ArrayAccess uses `$this->data` instead of `$this->_data`, and `$name` instead of `$offset` | `isset($ref['field'])` always returns false, set/unset trigger undefined variable warnings |
| Site tags | `frontend/init.inc.php` | `self::$head_tags` used outside class context | Fatal error when site tags with head/body position are loaded |

## Details

**MCP blocked_until:** Both MCP endpoints set `blocked_until` on failed login attempts (line 74) but never check it when loading the account. An attacker blocked at the normal login can continue brute-forcing via the MCP endpoint. Fix: add `and (blocked_until is null or blocked_until < NOW())` to the WHERE clause.

**SQL in exceptions:** `nod_database.inc.php` line 277 and 307 append `$sql` to the error message. Now the query is logged to `storage://logs/errors.log` with a reference ID, and only the error number + message + ref are in the exception.

**ArrayAccess:** Copy-paste bug — `offsetSet`/`offsetUnset` reference `$name` (from `__set`) instead of `$offset` (the ArrayAccess parameter). `offsetExists`/`offsetGet` use `$this->data` (public, doesn't exist) instead of `$this->_data` (protected, where the data lives). All `ref_*` classes inherit this.

**Site tags:** `self::` has no meaning inside an anonymous function that isn't bound to a class. Changed to `document::$head_tags` / `document::$foot_tags`.

## Test plan

- [ ] MCP login with blocked account returns 401 (not a successful auth)
- [ ] Database error messages no longer contain SQL queries
- [ ] `isset($ref_product['price'])` returns true for existing properties
- [ ] Site tags with position "head" or "body" render without fatal error
- [ ] All 34 platform tests still pass

### DIFF
--- a/public_html/backend/pages/mcp.inc.php
+++ b/public_html/backend/pages/mcp.inc.php
@@ -46,6 +46,7 @@
 				and lower(username) = lower('". database::input($_SERVER['PHP_AUTH_USER']) ."')
 				and (valid_from is null or valid_from < '". database::input(date('Y-m-d H:i:s')) ."')
 				and (valid_to is null or valid_to > '". database::input(date('Y-m-d H:i:s')) ."')
+				and (blocked_until is null or blocked_until < '". database::input(date('Y-m-d H:i:s')) ."')
 				limit 1;",
 			)->fetch();
 
--- a/public_html/frontend/init.inc.php
+++ b/public_html/frontend/init.inc.php
@@ -96,11 +96,11 @@
 		switch ($site_tag['position']) {
 
 			case 'head':
-				self::$head_tags[] = $site_tag['content'];
+				document::$head_tags[] = $site_tag['content'];
 				break;
 
 			case 'body':
-				self::$foot_tags[] = $site_tag['content'];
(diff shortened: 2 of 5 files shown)


timint:
Looks very good, I approve. 👍

Wachhund:
Thanks for the quick review!

PR #422, opened 2026-03-24
* Migrate database to InnoDB + DECIMAL, fix Docker auto-install

Wachhund:
MyISAM walked into a bar and ordered a utf8mb4. The bartender said "Sorry, your index can't handle that."

## Database migration

| Before | After |
|--------|-------|
| MyISAM (all tables) | InnoDB |
| FLOAT for monetary values | DECIMAL(12,4) |
| utf8mb4 installation broken (index key overflow) | utf8mb4 works out of the box |

### Why InnoDB?

- **utf8mb4 support**: MyISAM has a 1000-byte index key limit. With utf8mb4 (4 bytes/char), a `VARCHAR(256)` in a UNIQUE INDEX exceeds that limit. InnoDB DYNAMIC allows 3072 bytes — no problem.
- **Transactions**: InnoDB supports `BEGIN`/`COMMIT`/`ROLLBACK` — essential for order processing integrity.
- **Row-level locking**: Better concurrency for multi-user shops.
- **Crash recovery**: InnoDB has crash-safe storage; MyISAM doesn't.

### Why DECIMAL?

FLOAT has precision issues with monetary values (`0.1 + 0.2 != 0.3`). DECIMAL stores exact values — critical for prices, taxes, and totals.

### Upgrade path

`upgrade_patches/2.7.0.sql` converts all existing tables from MyISAM to InnoDB and FLOAT to DECIMAL. Non-destructive — no data loss.

## Docker fixes (included)

| Fix | Problem |
|-----|---------|
| CRLF in entrypoint | Windows line endings cause `exec: no such file or directory` in Linux container. Added `sed` strip in Dockerfile. |
| Installer CWD | `install.php` uses relative paths (`../includes/config.inc.php`). Entrypoint now `cd`s to install/ before running. |

## Test plan

- [x] Fresh Docker install with utf8mb4_general_ci completes successfully
- [x] Shop loads and functions after installation
- [ ] Existing v2 installation upgrades via 2.7.0.sql without errors
- [ ] All monetary values retain precision after FLOAT → DECIMAL conversion


timint:
I had issues with DECIMAL as mysqli passes it as a string to PHP and as of 8.x there are some functions throwing a warning or error when numerical strings are passed to functions that expects a non-string number. So instead of having to type convert arguments allover the platform I did FLOAT for consistent types with db.

2.x continued using MyISAM/Aria due to faster performance than InnoDB, but 3.0 will is InnoDB by default. 3.0 is also utf8mb4. I want to get 3.0 out as soon as possible so it's an option to postpone these changes.

Indeed relative paths in installer is problematic. But instead of changing working directory we can use __DIR__.

Let me hear your thoughts.

Wachhund:
Good points — let me address them one by one:

**DECIMAL as string:** You're right, mysqli returns DECIMAL as strings, and PHP 8.x is stricter about type juggling. `MYSQLI_OPT_INT_AND_FLOAT_NATIVE` doesn't help here either — it only casts INT and FLOAT, DECIMAL stays string by design (to preserve precision).

However, FLOAT has IEEE 754 rounding issues — `19.99` can become `19.9899997...` which is risky for monetary calculations in an e-commerce platform. I've actually dealt with this firsthand using Billomat for invoicing — weird rounded values showing up on invoices that were a real headache to debug. Turned out to be floating-point precision every time. Most of the math happens in SQL anyway (SUM, multiplication in queries), where DECIMAL handles it correctly. PHP-side, the values mostly flow back into queries as strings via `database::input()` or get formatted for display — both work fine with string representation.

For the few spots that need a numeric type in PHP, a simple `(float)` cast or `+ 0` does the job. That said, if you've already dealt with this across the codebase and prefer FLOAT for consistency, I understand the trade-off. Your call.

**InnoDB:** Good to hear 3.0 is already InnoDB — that aligns with this PR. We can drop the migration part if it's redundant.

**`__DIR__` instead of chdir:** Better approach, agreed. Cleaner than changing CWD.

**Timing:** Happy to postpone anything that could slow down the 3.0 release. The Docker fix (`__DIR__`) and InnoDB alignment seem safe, the DECIMAL change can wait if you prefer.

Wachhund:
By the way — if there's anything on your 3.0 roadmap where you could use an extra pair of hands, happy to help. Just point me at it.

timint:
Let's do this decimal implementation in 3.0 (dev-major) along with innodb and mb4. I'm afraid 2.7 might take a hit and it's better if we do that in 3.0. Floating points are a nightmare and if I can get your expertise in this I'm willing to make this move. Closing this PR for 2.x.

I don't have a public roadmap as it became a lot of work to keep up with. But it's something we should look into now that I have you just doing awesome things for the future of LiteCart. 🙂

Wachhund:
Makes total sense — keeping 2.7 stable and doing the DECIMAL + InnoDB + mb4 move properly in 3.0 is the right call.

Regarding a roadmap — even a lightweight one (a pinned issue or a simple markdown file) could go a long way. Would make it easier to coordinate and pick up tasks. Happy to help set something up if you want.

PR #423, opened 2026-03-24
* Replace FLOAT with DECIMAL for monetary and quantity columns

Wachhund:
FLOAT and money had a complicated relationship. DECIMAL is the couples therapist.

## The problem

All 51 monetary and quantity columns in v3 use FLOAT — an IEEE 754 binary type with inherent precision issues. Over thousands of order lines, this accumulates into real cent differences in totals, tax calculations, and payment reconciliation.

## The fix

Simple type change across two files:

| File | FLOAT before | DECIMAL after |
|------|-------------|--------------|
| structure.json | 51 columns | 51 columns |
| migrations/3.0.0.sql | 53 references | 53 references |

No upgrade patch needed — v3 has no production installations.

## Why DECIMAL for everything?

Quantities (quantity, quantity_min, quantity_max, quantity_step) benefit from exact arithmetic too — especially with fractional units (e.g. 0.5 kg, 1.5 liters). Weight and dimension columns follow the same logic.

Same change as v2 PR #422, adapted for v3.

## Test plan

- [ ] Fresh installation creates all tables with DECIMAL columns
- [ ] All 34 platform tests still pass
- [ ] Price calculations return exact values (no floating-point drift)


timint:
So now awaits anti-histamins for PHP 8's allergy towards numerical strings. 🫣😄

Wachhund:
Ha! I'll bring the medicine. Most spots should be fine since the values flow straight back into SQL as strings anyway. For the few places that need a numeric type, a `(float)` cast or `+ 0` does the trick. Happy to help track those down.

PR #424, opened 2026-03-24
+ Add transaction management to order processing, fix NOT IN data loss

Wachhund:
The database was playing Jenga with order data. One wrong move and half the blocks disappear.

## Transaction Management

`ent_order::save()` writes to 5 tables in sequence: `orders`, `orders_lines`, `orders_items`, `orders_comments`, and `stock_items`. Until now, each write ran in auto-commit — if the 3rd query failed, the first 2 were already committed. You'd end up with an order header but no lines, or decremented stock without an order.

Now the entire save is wrapped in `BEGIN`/`COMMIT`/`ROLLBACK`:

```php
database::begin_transaction();
try {
    // ... all writes ...
    database::commit();
} catch (Throwable $e) {
    database::rollback();
    throw $e;
}
```

Three new methods on the database node:

| Method | Wraps |
|--------|-------|
| `database::begin_transaction()` | `mysqli_begin_transaction()` |
| `database::commit()` | `mysqli_commit()` |
| `database::rollback()` | `mysqli_rollback()` |

These are thin wrappers — if we migrate to PDO later, only the internals change (3 lines), not the callers.

## NOT IN Data Loss Bug

Two DELETE queries in `save()` used this pattern:

```php
"and id not in ('". implode("', '", array_column($data, 'id')) ."')"
```

When the array is empty, this produces `id not in ('')` — which matches ALL rows (empty string cast to 0, no IDs are 0). Result: all order lines or comments silently deleted on next save.

Fixed by:
- Using `array_filter()` to remove empty IDs
- Using `intval()` cast instead of string quoting
- Skipping the NOT IN clause entirely when the array is empty

## Test plan

- [ ] All 34 platform tests pass
- [ ] Order with lines saves and reloads correctly
- [ ] Order save that triggers a DB error rolls back completely (no partial data)
- [ ] Saving an order with empty lines array doesn't delete existing lines

### DIFF
--- a/public_html/includes/entities/ent_order.inc.php
+++ b/public_html/includes/entities/ent_order.inc.php
@@ -174,6 +174,10 @@
 
 		public function save() {
 
+			database::begin_transaction();
+
+			try {
+
 			// Re-calculate total if there are changes
 			$this->refresh_total();
 
@@ -313,11 +317,12 @@
 				}
 			}
 
-			// Delete order lines
+			// Delete order lines not in current data
+			$line_ids = array_filter(array_column($this->data['lines'], 'id'));
 			database::query(
 				"delete from ". DB_TABLE_PREFIX ."orders_lines
 				where order_id = ". (int)$this->data['id'] ."
-				and id not in ('". implode("', '", array_column($this->data['lines'], 'id')) ."');"
+				". ($line_ids ? "and id not in (". implode(",", array_map('intval', $line_ids)) .")" : "") .";"
 			);
 
 			// Insert/update order lines
@@ -422,11 +427,12 @@
 				}
 			};
 
-			// Delete comments
+			// Delete comments not in current data
+			$comment_ids = array_filter(array_column($this->data['comments'], 'id'));
 			database::query(
 				"delete from ". DB_TABLE_PREFIX ."orders_comments
 				where order_id = ". (int)$this->data['id'] ."
-				and id not in ('". implode("', '", array_column($this->data['comments'], 'id')) ."');"
+				". ($comment_ids ? "and id not in (". implode(",", array_map('intval', $comment_ids)) .")" : "") .";"
 			);
 
 			// Insert/update comments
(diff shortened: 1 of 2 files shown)


timint:
I looked into PDO back in the days but it lacks some major features that mysqli has. Performance was also better with mysqli. I don't remember what PDO couldn't do.I think it was data seek and two more things.

Forcing strings to integers may turn "a" into 1, causing an id of 1 to be deleted. Let's think about that for a bit.

Transactions are useful. But they don't restore auto increment number or alterations. So we might need to add logic for that.

I'm still thinking some of these ideas should be postponed for 3.0.

Wachhund:
Good points, let me go through them:

**PDO:** Fully agree — we're not proposing PDO either. mysqli works well, no reason to change.

**intval concern:** Fair point, but `intval("a")` actually returns `0`, not `1` — and auto-increment IDs start at 1, so there's no risk of accidentally matching a real row. That said, the intval is just a secondary safeguard. The actual fix is the empty array guard — when no line IDs exist, the `NOT IN` clause is skipped entirely instead of producing `NOT IN ('')` which matches everything.

**Transactions + auto-increment gaps:** You're right that auto-increment doesn't roll back. Resetting it manually (`ALTER TABLE ... AUTO_INCREMENT`) would be risky in production — concurrent requests could grab an ID between rollback and reset, causing duplicates. But the gaps are harmless. They only occur when something actually goes wrong, and a gap in order numbers is far less problematic than a half-written order with missing lines or orphaned comments.

**Timing:** The NOT IN bug is a silent data loss risk that exists right now — I'd keep that fix regardless. The transaction wrapper is lightweight (three lines around existing code), but happy to split the PR if that helps.

timint:
Oh my bad, I thought intval("a") did what (int)$string did, I learn.
Innodb transactions welcome starting with 3.0 (dev-major). I actually use them already during tests, but rolled it back for failed upgrade migrations. I learned they are no solution when doing alterations.

Wachhund:
No worries — it's one of those PHP quirks that everyone assumes works differently until they check. The important part is the empty array guard, the intval is just belt-and-suspenders. Glad the transaction approach fits!

PR #426, opened 2026-03-24
* Fix 4 audit findings: XSS, preload, fetch_all, regex

Wachhund:
Four one-liner fixes found during a deep dive into the v3 code. All independent, all minimal risk.

## Fixes

| Fix | File | Bug | Impact |
|-----|------|-----|--------|
| XSS | `edit_product.inc.php:520,523,524,798` | `$_POST` values echoed without escaping | Reflected XSS in admin product editor |
| XSS | `edit_order.inc.php:977` | `nl2br($_POST[...])` without escaping | Reflected XSS in admin order comments |
| Preload | `nod_document.inc.php:484` | `$link` used instead of `$url` (undefined variable) | `add_preload()` silently broken — preload headers never sent |
| fetch_all | `nod_database.inc.php:333` | `$column=null, $index_column=null` in call overwrites parameters | `database::fetch_all($result, 'name')` ignores the column filter |
| Regex | `login.inc.php:163` | `preg_replace('[0-9]...')` uses brackets as delimiter | IP range check for 2FA always matches — potentially bypasses 2FA on unknown IPs |

## Test plan

- [ ] All 34 platform tests pass
- [ ] Product editor with attributes renders correctly (no broken HTML)
- [ ] Order comments display correctly
- [ ] `database::fetch_all($result, 'column_name')` returns only that column

### DIFF
--- a/public_html/backend/apps/catalog/edit_product.inc.php
+++ b/public_html/backend/apps/catalog/edit_product.inc.php
@@ -517,11 +517,11 @@
 									<?php echo f::form_input_hidden('attributes['.$key.'][value_id]', true); ?>
 									<?php echo f::form_input_hidden('attributes['.$key.'][value_name]', true); ?>
 									<?php echo f::form_input_hidden('attributes['.$key.'][custom_value]', true); ?>
-									<?php echo $_POST['attributes'][$key]['group_name']; ?>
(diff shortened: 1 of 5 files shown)


timint:
Lots of good stuff here 👍 You have the eyes of a hawk.

Wachhund:
Thanks! Glad they're useful. The deep dive keeps turning up little things — more to come.

PR #428, opened 2026-03-25
* Fix sitemap generator discarding product image data

Wachhund:
## Summary

The sitemap generator's "Product Images" section collected XML into a local `$output` array — that was never written anywhere. Every product image was silently discarded.

| What | Before | After |
|------|--------|-------|
| Product images in sitemap | Lost to the void | Written to file via `fwrite()` |
| `xmlns:image` namespace | Missing | Declared in `<urlset>` |
| Sitemap splitting | Skipped for images | Count/bump logic like all other sections |

## Changes

- **`$output[]` → `fwrite($fh, ...)`** — aligns with the pattern used by Categories, Products, and the `$draw_url_node` helper
- **Added `xmlns:image`** namespace to the `<urlset>` declaration — without it, `<image:image>` elements are invalid XML
- **Added count/bump logic** so product images respect the `entries_per_sitemap` setting

## Test plan

- [ ] Run sitemap generator job
- [ ] Verify product images appear in generated sitemap XML
- [ ] Verify `xmlns:image` namespace is present in `<urlset>`
- [ ] Verify sitemap splitting still works when entry count exceeds limit

### DIFF
--- a/public_html/includes/modules/jobs/job_sitemap_generator.inc.php
+++ b/public_html/includes/modules/jobs/job_sitemap_generator.inc.php
@@ -46,7 +46,7 @@
 				$fh = fopen($last_sitemap_path, 'x');
 				fwrite($fh, implode(PHP_EOL, [
 					'<?xml version="1.0" encoding="UTF-8"?>',
-					'<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:xhtml="http://www.w3.org/1999/xhtml">',
+					'<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:xhtml="http://www.w3.org/1999/xhtml" xmlns:image="http://www.google.com/schemas/sitemap-image/1.1">',
 				]) . PHP_EOL);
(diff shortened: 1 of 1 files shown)


timint:
Very thankful for this fix. 🙂

Wachhund:
Glad I could help! These silent bugs are the sneaky ones.

timint:
Oooh yes, I wish I had automated tests for everything. :)

Btw, I'm looking into using Github Projects for todos. I'm an old Trello user from before Atlassian ruined the product and I think this is the tool I want for this.

Wachhund:
GitHub Projects sounds great — I've never used it myself, been in Atlassian-land at my last job and currently looking at OneDev (running in Docker, quite promising actually). But any tool that brings structure to feature requests, status tracking, and design/architecture discussions would be a win.

As for test coverage — we're at ~78% for entities right now (28 of 36), plus the business logic tests for product catalog, stock, cart, and checkout. Nodes, references, and functions have no coverage yet, so there's plenty of room to grow. Happy to push that further when the time is right.

image

PR #430, opened 2026-03-27
+ Add unit tests for pure functions and unit conversions

Wachhund:
## Summary

First round of function and utility tests — covering the building blocks that everything else depends on.

| func_escape | XSS escaping (HTML, attr, JS, MySQL, LIKE) | 6 |
| func_validate | Email validation (valid + invalid addresses) | 11 |
| func_password | Generation (length, character classes, CSPRNG uniqueness) + strength check | 6 |
| func_string | string_split (delimiter, empty), string_ellipsis (truncate, short, empty, negative) | 5 |
| func_array | array_between, column_intact, each, update, intersect_compare, flatten/unflatten, depth, grep, diff_assoc_recursive | 10 |
| unit_conversion | Weight (kg/g/lb), Length (m/cm/ft/km/mi), Volume (L/gal/cm3) + unit definition integrity | 10 |

No bugs found this time — the pure utility functions are solid.

Side note: `array_exclude()` and `array_filter_recursive()` in func_array.inc.php have variable name bugs (`$input` instead of `$array`, `$v` instead of `$value`) but those are untested here to avoid PHP errors. Can address separately.

## Test plan

- [ ] CI green — all tests pass
- [ ] Verify no database dependencies in these tests (pure logic only)

### DIFF
--- a/tests/func_array.php
+++ b/tests/func_array.php
@@ -1,0 +1,144 @@
+<?php
+
+	include_once __DIR__.'/../public_html/includes/app_header.inc.php';
+
+	try {
+
+		########################################################################
+		## array_between
+		########################################################################
+
+		$result = f::array_between([1, 5, 10, 15, 20], 5, 16);
+
+		if (!in_array(5, $result) || !in_array(10, $result) || !in_array(15, $result)) {
+			throw new Exception('array_between failed to filter values in range');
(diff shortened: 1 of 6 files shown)


timint:
👍 Very useful, thank you

PR #433, opened 2026-03-27
+ Add phase 3 DB-dependent tests: settings, currency, country, format

Wachhund:
## Summary

Third round of tests — now covering subsystems that depend on database seed data.

| nod_settings | get/set/fallback, store_currency_code format, temporary override + restore | 5 |
| nod_currency | Loading, selection, calculate same-to-same, format_raw numeric, format with prefix/suffix, zero handling | 6 |
| ref_country | Load US/DE by ISO code, currency_code, zones array, invalid country fallback | 5 |
| func_format | format_json (pretty/compact), format_path_friendly (slugs, German umlauts), format_address with DB lookup, format_number | 6 |

Note: ref_country uses lazy loading via `__get` (property access). ArrayAccess `offsetGet()` does not trigger `_load()` — tests use `->` syntax accordingly.

## Test plan

- [x] CI green — 37/37 tests pass

### DIFF
--- a/tests/func_format.php
+++ b/tests/func_format.php
@@ -1,0 +1,114 @@
+<?php
+
+	include_once __DIR__.'/../public_html/includes/app_header.inc.php';
+
+	try {
+
+		########################################################################
+		## format_json — pretty print
+		########################################################################
+
+		$data = ['name' => 'Test', 'value' => 42];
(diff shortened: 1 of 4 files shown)


timint:
👍👍

PR #434, opened 2026-03-28
* Fix MySQL 8 compatibility, case-insensitive search, product page crash

Wachhund:
Pointed Docker at MySQL 8.0. It pointed back — with strict mode errors, case-sensitive searches, and a product page that divided by zero just to make a point.

### Case-insensitive Search

MySQL 8 `json_value()` returns `utf8mb4_bin` collation, making every LIKE comparison case-sensitive. Searching "sticker" on a German locale found nothing, while "Sticker" worked fine.

| What | MySQL 8 (before) | MySQL 8 (after) | MariaDB |
|------|-------------------|-----------------|---------|
| `json_value()` collation | `utf8mb4_bin` | forced `utf8mb4_unicode_ci` | already case-insensitive |
| `LIKE '%sticker%'` vs "Sticker" | no match | match | match |
| COLLATE clause effect | fixes it | — | harmless no-op |

**Fix:** Added `COLLATE utf8mb4_unicode_ci` to all 26 `json_value()...LIKE` patterns across frontend search, backend catalog, stock items, pages, and JSON endpoints.

### Product Page Crash

Controller set the snippet key as `default_image`, template expected `image`. Cascade: undefined variable → `is_file(null)` deprecated → `getimagesize()` returns false → division by zero in aspect ratio calculation.

| File | Issue | Fix |
|------|-------|-----|
| `frontend/pages/product.inc.php` | wrong snippet key `default_image` | renamed to `image` |
| `functions/func_draw.inc.php` | `is_file(null)` deprecated | guard for empty path |
| `entities/ent_image.inc.php` | array access on `getimagesize()` false | null-safe access |
| `functions/func_image.inc.php` | division by zero on empty aspect ratio | guard for empty divisor |

### Search Results Template

`$subcategories` variable was never set by the search controller — only `$categories`. Template iterated `$subcategories`, causing undefined variable + foreach-on-null warnings.

### Docker

Added `--character-set-client-handshake=FALSE` to MySQL config — forces all client connections to use the server's `utf8mb4` charset, preventing encoding mismatches from CLI tools.

## Changes

| Commit | Area | Description |
|--------|------|-------------|
| `58cd5d6` | Installer | MySQL 8 strict mode compatibility |
| `62eb64c` | Dockerfile | Use mysqli, remove built-in extensions |
| `d8587cf` | Runtime | Fix nullable TEXT column warnings |
| `8fe1320` | Search/Docker | Fix column name, category resolution, locale |
| `eb30955` | Search | Case-insensitive COLLATE + $subcategories fix |
| `4140a7a` | Product page | Fix $image undefined → division by zero |

## Test plan

- [x] Search (German locale): "sticker" lowercase finds products
- [x] Search (English locale): "sticker" lowercase finds products
- [x] Product page: no warnings, image renders correctly
- [x] Product without image: fallback image, no fatal error
- [x] Backend search: case-insensitive across products, categories, stock items, pages
- [x] Fresh `docker-compose up --build` with MySQL 8: installer + shop work
- [x] MariaDB: no regressions from COLLATE clause

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,14 +12,10 @@ RUN apt-get update && apt-get install -y \
     && docker-php-ext-configure gd --with-freetype --with-jpeg \
     && docker-php-ext-install -j$(nproc) \
         gd \
-        pdo_mysql \
+        mysqli \
         zip \
         intl \
-        curl \
         opcache \
-        dom \
-        simplexml \
-        mbstring \
     && apt-get clean && rm -rf /var/lib/apt/lists/*
 
 # Enable Apache modules
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,11 +5,14 @@ RUN apt-get update && apt-get install -y \
     libfreetype6-dev \
     libjpeg62-turbo-dev \
     libpng-dev \
+    libwebp-dev \
     libzip-dev \
     libicu-dev \
     libcurl4-openssl-dev \
     libxml2-dev \
-    && docker-php-ext-configure gd --with-freetype --with-jpeg \
+    locales \
+    && sed -i '/en_US.UTF-8/s/^# //' /etc/locale.gen && locale-gen \
+    && docker-php-ext-configure gd --with-freetype --with-jpeg --with-webp \
     && docker-php-ext-install -j$(nproc) \
         gd \
         mysqli \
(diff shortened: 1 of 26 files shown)


timint:
A complete candy mix 🍬🍫🙂 Thank you

Wachhund:
Thanks! Glad to deliver the full box 🍬 Took a little detour through the checkout while testing a payment module — turned out to be quite the adventure. Working on the findings from #435 next.

PR #439, opened 2026-04-01
[v3] * Fix syntax error in Data Studio import

Wachhund:
Spotted this one in CI — the new Data Studio is cool, just needs some parens.

## Summary

`new $class_name()->data` is a parse error on PHP < 8.4. Wrapping the instantiation in parentheses fixes it: `(new $class_name())->data`.

| File | Line | Fix |
|------|------|-----|
| `backend/apps/data_studio/import.inc.php` | 202 | `new $class_name()->data` → `(new $class_name())->data` |

## Test plan

- [x] CI passes (PHP lint)
- [x] Data Studio import still works on PHP 8.3

### DIFF
--- a/public_html/backend/apps/data_studio/import.inc.php
+++ b/public_html/backend/apps/data_studio/import.inc.php
@@ -199,7 +199,7 @@
 					];
 
 					$class_name = 'ent_' . $entity;
-					$properties = array_keys(f::array_flatten(new $class_name()->data));
+					$properties = array_keys(f::array_flatten((new $class_name())->data));
 
 					break;
 


Wachhund:
Data Studio and the Email app look really exciting — can't wait to take them for a spin once they're ready!

timint:
Yeah I got tired of writing content specific import scripts. I thought it's time I nail something universal that will work for everything.

It needs some work for sure, but I thought I should add it so it can be played with.

The most problematic thingy is CSV which is a two dimensional layer. Exporting entities with array properties could need serializing with json which is better than nothing.

PR #440, opened 2026-04-01
[v3] + Category-wide percentage discounts

Wachhund:
This one's been on the community wishlist for a while — merchants can now slap a discount on an entire category without touching individual products.

## Summary

| What | Details |
|------|---------|
| **New fields on categories** | `discount_percent` (0-100%), `discount_valid_from`, `discount_valid_to`, `discount_stacking` |
| **Price pipeline** | Applied after campaign + customer group prices as percentage reduction on `final_price` |
| **Multiple categories** | Highest active discount wins (`max()`) |
| **Validity** | Optional date range — NULL means always active. Expired/future discounts have no effect |
| **Frontend** | Strikethrough price + "-X%" badge on category listings and product pages |
| **Admin** | New "Discount" fieldset on category edit form |
| **Migration** | ALTER TABLE for existing installations (idempotent) |

### Files changed (9)

| Layer | File | Change |
|-------|------|--------|
| Schema | `install/structure.json` | 4 new columns on `categories` |
| Migration | `install/migrations/3.0.0.inc.php` | ALTER TABLE (idempotent) |
| Entity | `entities/ent_category.inc.php` | `save()` writes discount fields, server-side 0-100% clamp |
| Reference | `references/ref_category.inc.php` | Lazy property `active_discount` with validity check |
| Reference | `references/ref_product.inc.php` | `final_price` applies category discount; lazy property `category_discount` |
| Catalog query | `functions/func_catalog.inc.php` | JOIN category discounts, apply to `final_price` in SQL |
| Sticker | `functions/func_draw.inc.php` | "-X%" badge for category discounts |
| Product page | `frontend/pages/product.inc.php` | "-X%" sticker on product detail |
| Admin | `backend/apps/catalog/edit_category.inc.php` | Discount fieldset (percent, dates, stacking toggle) |

### Known limitations (v1)

- **Stacking with discount codes:** The `discount_stacking` field is stored but not enforced at checkout — discount codes are an add-on, so integration depends on the add-on's API. The field is ready for when that integration makes sense.
- **Order line tracking:** The discounted price is saved correctly in orders, but the discount source (category vs campaign) is not yet written to `orders_lines.userdata`. Planned for a follow-up.
- **Campaign + category:** Both discounts compound (category % applied on top of campaign price). This is more generous to customers but could give unexpectedly deep discounts.

### Use case

A merchant wants to run a Black Friday sale on their own products (grouped in a specific category), while excluding dealer products in other categories. No coupon code needed — the discount applies automatically with a visible "-20%" badge.

**Source:** Community forum request ([tomydevon, Nov 2025](https://www.litecart.net/de/forums/5/feature-requests-and-new-ideas/24435/discount-on-an-entire-product-category-rather-than-on-individual)). Tim noted it's been requested multiple times.

## Test plan

- [ ] Create a category, set 20% discount, save
- [ ] Verify products in that category show strikethrough price + "-20%" badge on category page
- [ ] Verify same display on product detail page
- [ ] Add to cart — discounted price used as line item price
- [ ] Set `valid_from` in the future — discount should NOT apply
- [ ] Set `valid_to` in the past — discount should NOT apply
- [ ] Set discount to 0% — products show regular price
- [ ] Product in two categories with different discounts — highest applies
- [ ] Products in subcategories are NOT affected by parent's discount
- [ ] Existing campaign products still show campaign sticker when no category discount

### DIFF
--- a/public_html/backend/apps/catalog/edit_category.inc.php
+++ b/public_html/backend/apps/catalog/edit_category.inc.php
@@ -64,6 +64,10 @@
 				'synonyms',
 				'filters',
 				'priority',
+				'discount_percent',
+				'discount_valid_from',
+				'discount_valid_to',
+				'discount_stacking',
 			] as $field) {
 				if (isset($_POST[$field])) {
 					$category->data[$field] = $_POST[$field];
@@ -178,6 +182,35 @@
 								<div class="form-label"><?php echo t('title_priority', 'Priority'); ?></div>
 								<?php echo f::form_input_number('priority', true); ?>
 							</label>
+
+							<fieldset style="margin-top: 15px; padding: 15px; border: 1px solid var(--border-color); border-radius: 5px;">
+								<legend style="font-weight: bold; padding: 0 5px;"><?php echo t('title_discount', 'Discount'); ?></legend>
+
+								<label class="form-group">
+									<div class="form-label"><?php echo t('title_discount_percent', 'Discount (%)'); ?></div>
+									<?php echo f::form_input_number('discount_percent', true, 'min="0" max="100" step="0.01"'); ?>
+								</label>
+
+								<div class="grid">
+									<div class="col-md-6">
+										<label class="form-group">
+											<div class="form-label"><?php echo t('title_valid_from', 'Valid From'); ?></div>
+											<?php echo f::form_input_datetime('discount_valid_from', true); ?>
+										</label>
+									</div>
+									<div class="col-md-6">
+										<label class="form-group">
+											<div class="form-label"><?php echo t('title_valid_to', 'Valid To'); ?></div>
+											<?php echo f::form_input_datetime('discount_valid_to', true); ?>
+										</label>
+									</div>
+								</div>
+
+								<label class="form-group">
+									<div class="form-label"><?php echo t('title_discount_stacking', 'Stackable With Discount Codes'); ?></div>
+									<?php echo f::form_toggle('discount_stacking', 'y/n', true); ?>
+								</label>
+							</fieldset>
 						</div>
 
 						<div class="col-md-4">
--- a/public_html/frontend/pages/product.inc.php
+++ b/public_html/frontend/pages/product.inc.php
@@ -232,7 +232,9 @@
 	}
 
 	// Sticker
-	if (!empty($product->campaign['price']) && $product->price > 0 && $product->campaign['price'] > 0) {
+	if ($product->category_discount > 0) {
+		$_page->snippets['sticker'] = '<div class="sticker sale">-'. (int)$product->category_discount .'%</div>';
+	} else if (!empty($product->campaign['price']) && $product->price > 0 && $product->campaign['price'] > 0) {
 		$percentage = round(($product->price - $product->campaign['price']) / $product->price * 100);
 		$_page->snippets['sticker'] = '<div class="sticker sale">'. t('sticker_sale', 'Sale') .' -'. $percentage .'%</div>';
 	} else if ($product->created_at > date('Y-m-d', strtotime('-'.settings::get('new_products_max_age')))) {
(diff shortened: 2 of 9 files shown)


timint:
I like this idea very much but have to think about it for a while. The reason is I want to keep the pricing and campaign management as consistent as possible in the platform. Meaning this might go better as an add-on, or it might need some adjustments. Spontaneously my idea is it could be set inside a campaign rather than in category settings. I will let the thoughts mature a bit.

Off-topic: How are your UI skills? I haven't come to a point where I feel the frontend is ready for an audience yet. I think it still needs polishing to give a more professional look.

Wachhund:
Take your time — getting the pricing model right is more important than shipping fast.

A quick thought on the add-on route: category discounts touch quite a few core layers — price pipeline (ref_product final_price), catalog SQL queries (JOIN for bulk listings), frontend stickers, and order storage. Hooking all of that cleanly from an add-on through vMod would be fairly complex and fragile across updates. A core integration (whether in category settings or inside the campaign system) would be more robust.

That said, both paths are possible. The campaign approach actually sounds appealing — campaigns already have date ranges and price overrides, so extending them with a "category scope" could keep the pricing logic in one place. Happy to rework the PR in that direction if you settle on it.

Wachhund:
On the UI question — yes, actually! I had some ideas and experimented a bit with the v2 frontend a while back, but once the focus shifted to v3 I shelved that.

I'd be happy to take a closer look at the v3 frontend and put together a design draft. Any specific areas you'd like me to focus on, or particular ideas you have in mind?

timint:
> A quick thought on the add-on route: category discounts touch quite a few core layers — price pipeline (ref_product final_price), catalog SQL queries (JOIN for bulk listings), frontend stickers, and order storage. Hooking all of that cleanly from an add-on through vMod would be fairly complex and fragile across updates. A core integration (whether in category settings or inside the campaign system) would be more robust.

Indeed, and the price pipeline needs to be as clean and consitent as possible. I am leaning towards the campaign layer. That way we can do the same for brands in one same place. What would have been nice is one central component for all price pipelining. As of now it's mainly a subquery that is joined with a products query. I tried to keep it central with func_catalog and ref_product. But every now and then there is a need to pipe of prices again separately.

>I'd be happy to take a closer look at the v3 frontend and put together a design draft. Any specific areas you'd like me to focus on, or particular ideas you have in mind?

This is music to my ears. We programmers try our best in UI but it's rare that we score the highest in terms of design. I like the look of many Tailwind based frameworks. But my goal has always been to stay dependency free, and superlightweight. With the establishment of [LiteCore ](https://github.com/litecore/framework) (the naked base of LiteCart) I am trying to keep  it clean, consistent, reusable, and easy accessible lightweight core components.

I like this design
image

Wachhund:
The campaign layer makes a lot of sense — categories and brands in one unified place is cleaner than scattered discount logic. And a central price pipeline component would be the ideal long-term target. We can shelve this PR and revisit the approach when the campaign system is ready for it.

On the UI front — love that direction. LiteCore is a solid foundation, and keeping it dependency-free is the right call. I've been looking at the framework source and the LESS architecture is well-structured for theming.

I'll put together a design draft that builds on LiteCore's existing components — no external libs, just better use of what's already there. The dark theme with gradient accents you shared is a great reference point. Will focus on: typography, spacing, card treatments, and making the catalog feel more premium while staying lightweight.

One note: as I work through the frontend, I may find small improvements for LiteCore components. Would it make sense to submit those as separate PRs to [litecore/framework](https://github.com/litecore/framework)?

Wachhund:
Here's a first design direction for the storefront polish — all built on LiteCore, zero dependencies.

**Typography:** Bump from 14px/1.2 to 15px/1.5 — noticeably more readable, WCAG compliant.

**Light theme tweaks:**
- White navbar instead of gradient — lets products be the focus
- Dark footer (`#1a1a2e`) — grounds the page, common pattern in modern shops
- Softer shadows, slightly larger border-radius (12px)

**Dark mode** (complete palette ready, following the backend pattern):
- Deep navy backgrounds (`#0f0f17`, `#1a1a2e`)
- Gold/amber accent (`#d4a853`) for stickers and highlights — inspired by your reference screenshot
- All via CSS custom properties, cookie-based toggle like the backend

**Product cards:** Subtle hover lift (`translateY` + shadow) instead of border-only. "Add to Cart" visible on mobile.

**Checkout:** Replace the infinite red flash with a solid accent color. Cleaner option card selection.

Each change would be a separate PR. Happy to adjust direction before implementing — these are proposals, not decisions.

PR #445, opened 2026-04-01
[v3] + Add title_person_name translation key (fixes #82)

### LINKED ISSUE #82: Title_name translation issue
> Small issue with "Name" translation in Lithuanian language (possibly other too)
> name for people is "vardas" and name for places or things is "pavadinimas"..
> something similar to "title_customer_name" should be added

Wachhund:
Nine years in the making — finally a clean fix for the Lithuanian name translation issue.

## Summary

Languages like Lithuanian, Czech, and Arabic use different words for "Name" depending on whether it refers to a person or a thing. The existing `title_name` key is used in 57 places for both contexts.

This PR introduces `title_person_name` for the 4 locations where "Name" refers to a person:

| File | Context |
|------|---------|
| `administrators/administrators.inc.php` | Admin listing table |
| `customers/customers.inc.php` | Customer listing table |
| `customers/customer_picker.inc.php` | Customer picker modal |
| `customers/newsletter_recipients.inc.php` | Newsletter recipients table |

The remaining 53 uses of `title_name` (products, categories, brands, modules, etc.) stay unchanged.

**Backwards-compatible:** The default is `"Name"` — identical to before. Languages that don't translate `title_person_name` see no change. Languages that need the distinction can now provide a separate translation.

## Test plan

- [ ] Admin list, customer list, customer picker, newsletter recipients all show "Name" column header (no visual change in English)
- [ ] All other admin pages still show "Name" for product/category/brand columns
- [ ] Translation system registers `title_person_name` as a new translatable key

### DIFF
--- a/public_html/backend/apps/administrators/administrators.inc.php
+++ b/public_html/backend/apps/administrators/administrators.inc.php
@@ -90,7 +90,7 @@
 					<th></th>
 					<th></th>
 					<th><?php echo t('title_username', 'Username'); ?></th>
-					<th><?php echo t('title_name', 'Name'); ?></th>
+					<th><?php echo t('title_person_name', 'Name'); ?></th>
 					<th class="main"><?php echo t('title_email', 'Email'); ?></th>
 					<th><?php echo t('title_restrictions', 'Restrictions'); ?></th>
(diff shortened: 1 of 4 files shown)


timint:
Although I never liked this one, I feel there is no better way around it. We just have to learn to like it 🙂

PR #446, opened 2026-04-02
[v3] + Campaign scope discounts for categories and brands

Wachhund:
Following up on the discussion in #440 — this moves category discounts into the campaign system, exactly as you suggested. Same mechanism works for brands too.

## Summary

Extends the existing campaign system with a **percentage discount mode** that targets entire categories and/or brands. Existing fixed-price campaigns are completely untouched.

| What | Details |
|------|---------|
| **New column on campaigns** | `discount_mode` (fixed/percentage), `discount_percent` (0-100%) |
| **New table** | `campaigns_scopes` — links campaign to category IDs and/or brand IDs |
| **Admin UI** | Mode toggle on campaign edit: "Fixed Prices" (existing) or "Percentage Discount" (new with category/brand pickers) |
| **Campaign list** | Shows type column ("-20%" or "Fixed Prices") and affected product count |
| **Price pipeline** | Scope percentage applied after regular/campaign-fixed/customer-group pricing. Lowest resulting price wins when both fixed and percentage campaigns apply |
| **Frontend** | "-X%" badge on category and product pages. Strikethrough price via existing draw_price_tag |
| **Migration** | ALTER TABLE + CREATE TABLE (idempotent) |

### How it works

A merchant creates a campaign, selects "Percentage Discount" mode, sets 20%, picks one or more categories and/or brands. All products in those categories or brands automatically get the discount — no need to add products individually. The campaign's existing date range and status controls apply.

### Files changed (9)

| Layer | File | Change |
|-------|------|--------|
| Schema | `structure.json` | 2 new columns on campaigns, new campaigns_scopes table |
| Migration | `migrations/3.0.0.inc.php` | ALTER + CREATE (idempotent) |
| Entity | `ent_campaign.inc.php` | Load/save scopes, discount_mode, discount_percent |
| Reference | `ref_product.inc.php` | Scope discount in campaign property + final_price pipeline |
| Catalog | `func_catalog.inc.php` | UNION-based scope discount JOIN in bulk query |
| Sticker | `func_draw.inc.php` | "-X%" badge for scope campaigns |
| Product page | `frontend/pages/product.inc.php` | Scope sticker on product detail |
| Admin edit | `edit_campaign.inc.php` | Mode toggle, percent field, category/brand pickers |
| Admin list | `campaigns.inc.php` | Type column, scope product count |

### Relationship to #440

This replaces the approach in #440 (discount columns on categories table). The campaign-based approach is cleaner because:
- Categories and brands use the same mechanism
- Campaigns already have date ranges, status, customer groups
- Price management stays centralized in one place

#440 can be closed once this is merged.

## Test plan

- [ ] Create a percentage campaign targeting a category — products show "-X%" badge and discounted price
- [ ] Create a percentage campaign targeting a brand — same behavior
- [ ] Combine category + brand in one campaign — union of both scopes
- [ ] Existing fixed-price campaigns still work unchanged
- [ ] Campaign list shows type and affected product count
- [ ] Expired/disabled campaign — no discount applied
- [ ] Product in multiple scope campaigns — highest percentage wins
- [ ] Product with both fixed campaign price and scope percentage — lower price wins
- [ ] Mode toggle in admin switches between product table and scope pickers

### DIFF
--- a/public_html/backend/apps/catalog/campaigns.inc.php
+++ b/public_html/backend/apps/catalog/campaigns.inc.php
@@ -34,7 +34,7 @@
 
 	// Table Rows, Total Number of Rows, Total Number of Pages
 	$campaigns = database::query(
-		"select c.*, cp.num_products
+		"select c.*, cp.num_products, csp.num_scope_products
 		from ". DB_TABLE_PREFIX ."campaigns c
 		left join (
 			select campaign_id, count(*) as num_products
@@ -42,6 +42,14 @@
 			where campaign_id is not null
 			group by campaign_id
 		) cp on (cp.campaign_id = c.id)
+		left join (
+			select cs.campaign_id, count(distinct p.id) as num_scope_products
+			from ". DB_TABLE_PREFIX ."campaigns_scopes cs
+			left join ". DB_TABLE_PREFIX ."products_to_categories ptc on (cs.scope_type = 'category' and cs.scope_id = ptc.category_id)
+			left join ". DB_TABLE_PREFIX ."products p on (p.id = ptc.product_id or (cs.scope_type = 'brand' and p.brand_id = cs.scope_id))
+			where p.id is not null
+			group by cs.campaign_id
+		) csp on (csp.campaign_id = c.id)
 		order by c.status desc, c.valid_from, c.valid_to;"
 	)->fetch_page(null, null, $_GET['page'], null, $num_rows, $num_pages);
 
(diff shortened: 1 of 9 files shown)


timint:
structure.sql keeps taking hits. See if you have some formatter that keeps re-rendering the file. It's a ping pong game 😄

timint:
In general this looks very good. I'd be happy to accept it and try it out.

timint:
We might not need the migrations added in 3.0.0.inc.php. As the upgrade script will patch the database using structure.json with all missing tables, column and indexes. It's not capable of doing renaming, moving or inserting data so those are the crucial parts of the migration patches. This is all new in 3.0. I thouhgt I might let you know of these new capabilites. 🙂

Wachhund:
Fixed both in the latest push:

1. **structure.json** — restored your format with spaces. The issue was my Node.js script stripping whitespace when rebuilding the file. I now edit the upstream file directly with targeted insertions instead of regenerating. Should be the last ping pong round.

2. **Migration removed** — good to know about the auto-patcher! That's a great capability for 3.0. Removed the migration entries, structure.json is the only source of truth now.

timint:
Very well done 👍

Wachhund:
Thanks Tim! Looking forward to seeing it in action. 🎯

timint:
I might throw in some changes to it later. I see a possibility of adding different percentage for different categories and brands. Picking them like we do products. That might require a category and brand picker modal/widget first. I have some tech, and might need to throw in some more for doing that.

PR #451, opened 2026-04-05
[v3] * Fix console errors on fresh install (fixes #448)

### LINKED ISSUE #448: [v3] Console errors on fresh install: privacyConsent and keep_alive 404
Found two JS console errors on a fresh v3 installation with demo data (MySQL 8.0, Standard install mode):

## Errors

**1. `Could not initiate privacy consent manager: $(...).privacyConsent is not a function`**

The privacy consent JS component is referenced but not loaded. Likely a missing script include or a dependency that isn't bundled in the Standard install mode.

**2. `GET /ajax/keep_alive 404 (Not Found)`**

The session keep-alive AJAX endpoint doesn't exist. The request is triggered periodically by the frontend JS.

## Reproduction

1. Fresh install with demo data on MySQL 8.0
2. Open browser console on the storefront homepage
3. Both errors appear immediately

## Environment

- PHP 8.3.30, MySQL 8.0.45, Apache 2 (Docker)
- Standard install mode (not Advanced)
- No vMods or add-ons installed

Wachhund:
Fresh install, open console, two red lines staring back at you. Not the best first impression.

## Fixes

| Error | Cause | Fix |
|-------|-------|-----|
| `privacyConsent is not a function` | Inline script in `site_privacy_consent.inc.php` runs before `app.js` is loaded (script order in `default.inc.php`) | Wrap in `waitFor('jQuery')` — same pattern used everywhere else |
| `GET /ajax/keep_alive 404` | LiteCore's `base.js` calls this endpoint every 60s but the route doesn't exist | Add `frontend/pages/ajax/keep_alive.inc.php` — simple JSON response to keep the session alive |

## Test plan

- [ ] Fresh install with demo data — no console errors on homepage
- [ ] Privacy consent banner initializes correctly
- [ ] Network tab shows `/ajax/keep_alive` returning 200 with `{"status":"ok"}`
- [ ] Session stays alive after 60+ seconds of inactivity

### DIFF
--- a/public_html/frontend/pages/ajax/keep_alive.inc.php
+++ b/public_html/frontend/pages/ajax/keep_alive.inc.php
@@ -1,0 +1,9 @@
+<?php
+
+  header('X-Robots-Tag: noindex');
+  header('Content-Type: application/json; charset=UTF-8');
+  header('Cache-Control: no-store');
+
+  echo json_encode(['status' => 'ok']);
+
+  exit;
--- a/public_html/frontend/templates/default/partials/site_privacy_consent.inc.php
+++ b/public_html/frontend/templates/default/partials/site_privacy_consent.inc.php
@@ -76,11 +76,13 @@
 </div>
 
 <script>
-	try {
-		const privacy_classes = <?php echo f::format_json($privacy_classes, ''); ?>;
-		const consents = <?php echo f::format_json($consents, ''); ?>;
-		$('#site-privacy-consent').privacyConsent(privacy_classes, consents);
-	} catch (e) {
-		console.error('Could not initiate privacy consent manager:' + e.message);
-	};
+	waitFor('jQuery', function($) {
+		try {
+			var privacy_classes = <?php echo f::format_json($privacy_classes, ''); ?>;
+			var consents = <?php echo f::format_json($consents, ''); ?>;
+			$('#site-privacy-consent').privacyConsent(privacy_classes, consents);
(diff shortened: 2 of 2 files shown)


timint:
Another idea can be doing this where it is output to ensure all on page scripting awaits jquery.
https://github.com/litecart/litecart/blob/dev-major/public_html/includes/nodes/nod_document.inc.php#L203
https://github.com/litecart/litecart/blob/dev-major/public_html/includes/nodes/nod_document.inc.php#L275-L277

Wachhund:
Good idea! I took a closer look at the scope:

The `waitFor('jQuery')` wrapper in `nod_document.inc.php` (line 264-267) wraps all **inline scripts** that get extracted and reinserted before `</body>`. If jQuery's `<script src>` tag is placed right before that inline block, the wrapper becomes unnecessary — jQuery is guaranteed to be loaded by source order.

However, the **28 external component files** (framework.js, app.js) each have their own `waitFor('jQuery')` wrapper and those should stay — their load order relative to jQuery isn't guaranteed unless we enforce it via `defer` or explicit ordering.

So the change would be scoped to `nod_document.inc.php` only:
1. Extract the jQuery `<script src>` from external scripts and insert it right before the inline script block
2. Replace the `waitFor('jQuery', function($) { ... })` wrapper with a simple IIFE `(($) => { ... })(jQuery)`

The component files remain untouched. Want me to put together a PR for this?

PR #452, opened 2026-04-05
[v3] * Replace var_dump with var_export in console_dump()

Wachhund:
## Summary
- Replace `var_dump()` with `var_export($vars, true)` in `console_dump()`
- Eliminates the `ob_start()`/`ob_get_clean()` workaround since `var_export` supports a return flag
- Fixes the CI "debug statement found" annotation on unchanged files

## Test plan
- [ ] Call `console_dump(['test' => 123, 'active' => true])` and verify output in browser console
- [ ] Confirm CI linting passes without `var_dump` annotation

### DIFF
--- a/public_html/includes/functions.inc.php
+++ b/public_html/includes/functions.inc.php
@@ -3,9 +3,7 @@
 	// Output any variable to the browser console
 	function console_dump(...$vars) { // ... as of PHP 5.6
 
-		ob_start();
-		var_dump($vars);
-		$output = ob_get_clean();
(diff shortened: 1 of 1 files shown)


timint:
var_export has a slightly different output and does not provide information about the object lengths. I'm positive to this change, but I can also consider JSON serializing it for enabling the browser console to decode the object.

Wachhund:
Good point! JSON encoding is definitely the better approach — the browser console can natively parse and display interactive objects. Follow-up PR: #455

PR #449, opened 2026-04-05
[v3] * Polish storefront visual appearance

Wachhund:
Spent the evening tweaking pixels and staring at rubber ducks until they looked premium. Here's what came out of it.

## Summary

| Change | Details |
|--------|---------|
| **Typography** | 15px base size (was 14px), 1.5 line-height (was 1.2) — noticeably more readable |
| **Navigation** | White background instead of gradient, subtler shadow |
| **Cards** | 12px border-radius (was 10px), lighter shadows, visible border on product cards, hover lift effect (translateY + shadow) |
| **Product names** | Line-clamp(2) instead of fixed 40px height — handles long names gracefully |
| **Buttons** | Standard green (#4caf50) for success/buy, more saturated primary blue |
| **Footer** | Dark background (#1a1a2e) with light text, ghost-style social buttons with border, dark copyright area |
| **Footer HTML** | Removed duplicate col-md-3 on inner section (was causing overflow) |
| **Checkout** | Removed infinite red flash animations on submit button and required checkboxes |
| **Scroll-up** | Styled as circular button with background, visible on dark footer |
| **Momentum scroll** | Added overflow:hidden to prevent horizontal page overflow |
| **Container** | max-width: 90% for breathing room on all screen sizes |

All changes are CSS custom property value adjustments + minor LESS rule tweaks. No new dependencies.

### Files changed (11)

| File | Change |
|------|--------|
| `less/variables.less` | Updated ~20 variable values (typography, colors, footer, buttons) |
| `less/app.less` | Product card hover lift, line-clamp, footer styling, scroll-up button, momentum scroll overflow, container max-width |
| `less/checkout.less` | Removed flasher animations |
| `partials/site_footer.inc.php` | Removed duplicate column classes on inner section |
| `css/*` | Compiled CSS (5 files) |

## Test plan

- [ ] Homepage: white navbar, rounded cards with visible borders, product hover lift works
- [ ] Product names: long names truncate with ellipsis instead of clipping
- [ ] Footer: dark background, light text, ghost social buttons, facility image fits in column
- [ ] Mobile (375px): no horizontal overflow, cards have breathing room
- [ ] Checkout: submit button has bold red text but no flashing animation
- [ ] Success/buy buttons are green (#4caf50)
- [ ] Scroll-to-top button visible on dark footer
- [ ] Product carousels don't cause horizontal page overflow

PR #457, opened 2026-04-05
[v3] * UI polish follow-up: container max-width, listing hover, footer tone

Wachhund:
Follow-up to #449 based on your feedback.

## Summary
- **Container max-width**: `max-width: min(90%, 1600px)` — stays fluid below 1600px, then caps out for ultrawide displays
- **Listing hover clipping**: Removed `overflow: hidden` from `.product` — it was clipping the box-shadow on hover when `translateY(-2px)` pushed the element up. The `.image-wrapper` inside already has its own `overflow: hidden` for the zoom effect
- **Footer tone**: Reduced blue saturation from 28% to 12% (`#1a1a2e` → `#1e1e26`, copyright `#151525` → `#1a1a22`)
- Compiled CSS and source maps included

## Test plan
- [ ] Verify container doesn't stretch beyond 1600px on ultrawide displays
- [ ] Hover over product listings — box-shadow should be fully visible (no clipping at top)
- [ ] Check footer and copyright bar appear more neutral/less blue


timint:
Sorry, I had some polishing that I pushed and broke this PR.
The 90% is problematic. With the floating 1440 standard I have a the most beautiful pixel placeholder standard on the front page. Let's give the 1600 standard some more thought.
image
10% dynamic spacing instead of fixed spacing I gotta think that one through also.

Wachhund:
Closing this — @timint addressed the container max-width and footer tone in his recent commits. The only remaining item (product listing overflow: hidden clipping box-shadow on hover) will be resubmitted as a smaller, focused PR.

PR #462, opened 2026-04-09
* Fix undefined variables and type errors found by static analysis

Wachhund:
Gave the codebase a mass X-ray — turns out some variables were living double lives under fake names, and a few never existed at all.

Ran Qodana against the codebase and found ~30 real bugs hiding behind 5000+ false positives. Here are the ones that actually bite at runtime.

## Summary

- **15 files** fixed across entities, form functions, modules, infrastructure, and tests
- All fixes are single-line variable corrections — no logic changes, no refactoring
- Found by JetBrains Qodana static analysis (PhpUndefinedVariableInspection)

### Entities

| File | Fix |
|------|-----|
| `ent_category` | `$image->type` accessed before instantiation in `save_image()`; `$id` undefined in `delete()` |
| `ent_order` | Luhn check-digit used undefined `$stack`; `_generate_order_number()` missing return; operator precedence bug in length digit |
| `ent_image` | `$filename` undefined in EXIF rotation (should be `$this->_file`); `load_from_string()` missing return |
| `ent_database_table_row` | `$row` undefined in `save()` (should be `$this->data`); `array_walk` returns bool, replaced with `array_map`; `_primary_column_name` vs `_primary_column` mismatch |

### Form Functions

| File | Fix |
|------|-----|
| `func_form` | `form_toggle()` deprecation handler swapped wrong variables; `form_select_function()` missing `$input` param + invalid early return; `form_select_multiple_files()` used `$options` instead of `$pattern`; `escape_js()` called with undefined `$dropdown`; `form_button()` fallback used undefined `$value` |

### Infrastructure

| File | Fix |
|------|-----|
| `mod_checkout` | `$module_id` / `$row` undefined in `options()` |
| `app_footer` | Operator precedence: `!$x = get() \|\| ...` always evaluates to true |
| `verify_checkout` | Logged `$result['error']` instead of `$error_message` |
| `download.inc` | `$file` undefined, `$folder` was set instead |
| `edit_cell` | `$value` undefined, should be `$_POST['value']` |
| `newsletter_recipients` | `$recipient_id` instead of `$recipient` in duplicate check |
| `product_stock_options` | `$notice` undefined when stock is available |
| `about.inc` | `$logfile` typo, should be `$log_file` |
| `upgrade.php` | `catch(Throwable $t)` but used `$e->getMessage()` |

### Tests

| File | Fix |
|------|-----|
| `func_password` | Test string `'Str0ng!Pass'` (11 chars) too short for 12-char minimum in `password_check_strength()` |

## Test plan

- [x] PHP lint passes on all 15 files
- [x] CI pipeline passes (lint, build, platform tests)
- [ ] Checkout flow completes without errors
- [ ] Background jobs trigger correctly (app_footer fix)
- [ ] Category image upload works
- [ ] Order number generation produces valid check digits
- [ ] Password strength test passes

### DIFF
--- a/public_html/backend/apps/customers/newsletter_recipients.inc.php
+++ b/public_html/backend/apps/customers/newsletter_recipients.inc.php
@@ -26,7 +26,7 @@
 				if (database::query(
 					"select *, concat(firstname, ' ', lastname) as name
 					from ". DB_TABLE_PREFIX ."newsletter_recipients
-					where email = '". database::input(strtolower($recipient_id)) ."'
+					where email = '". database::input(strtolower($recipient)) ."'
 					limit 1;"
 				)->num_rows) {
 					$newsletter_recipient = new ent_newsletter_recipient($recipient);
--- a/public_html/backend/apps/data_studio/edit_cell.inc.php
+++ b/public_html/backend/apps/data_studio/edit_cell.inc.php
@@ -42,6 +42,7 @@
 		}
 
 		// Determine SQL value and validation
+		$value = $_POST['value'] ?? '';
 		$is_null = false;
 		if ($value === '' && $column_info['null']) {
 			$is_null = true;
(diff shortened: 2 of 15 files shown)


timint:
Perfection! 🙂

Wachhund:
Thanks Tim, glad it's useful! 🙂

PR #474, opened 2026-04-17
[v3] Harden installer & upgrader against unauth remote overwrite

Wachhund:
The install.lock existed. install.php just never bothered to ask.

## What changes

| File | Change |
|---|---|
| `install/install.php` | Lock-check before any destructive step; var_export-based config write; shared bootstrap |
| `install/upgrade.php` | Admin-session gate for web path; CLI stays open |
| `install/index.php` | Uses shared bootstrap + headers helper |
| `install/config` | Placeholders always in `'…'` tokens (no trailing `/` inside the literal) |
| `install/includes/header.inc.php` | CSP nonce on inline `<script>`/`<style>` |
| `install/includes/bootstrap.inc.php` | New — CLI detection, FS constants, lock check |
| `install/includes/security_headers.inc.php` | New — CSP + nosniff + XFO + Referrer + HSTS |
| `install/includes/config_writer.inc.php` | New — `var_export()`-based config generation |
| `tests/install_access_control.php` | New — 5 suites covering the new helpers |

## Why

- `POST /install/install.php?install=1` on a deployed shop used to overwrite `storage/config.inc.php`, drop tables for the given prefix, and bootstrap an attacker-owned admin — even though `install.lock` was present. The lock was only checked in `index.php`. Fixed.
- `client_ip` from the request flowed through `strtr()` straight into a PHP single-quoted literal in the generated config, so a payload like `x']) || system($_GET['cmd']) || (['` wrote executable PHP to `storage/config.inc.php`. Replaced with `var_export()`.
- `upgrade.php?upgrade=true` ran ALTER TABLE / backups without any auth check. Added an administrator session gate; CLI stays exempt (`php_sapi_name() === 'cli'` + `$argv` present).
- Installer/upgrader pages shipped no security headers at all. Added CSP with per-request nonce, nosniff, XFO DENY, Referrer-Policy, HSTS behind proxy-aware HTTPS detection.

## Test plan

- [x] `tests/install_access_control.php` — 5 suites, all green
- [x] `php -l` across 78 files in `install/` — no syntax errors
- [x] Config-injection resistance verified by running `php -l` on output generated from 8 malicious payloads (embedded quotes, CRLF, backslash)
- [x] `client_ip` validator: IPv4, IPv6, invalid, CRLF, empty
- [x] Lock detection covers present/absent states
- [x] HTTPS detection honours `X-Forwarded-Proto` and `X-Forwarded-SSL`
- [ ] Browser smoke: `install.php` + pre-existing lock → 403 (staging)
- [ ] Browser smoke: `upgrade.php?upgrade=true` without session → 401 (staging)

## Rollback

Single squash commit; `git revert <sha>` restores the prior entry points. No DB migrations, no data changes.

## Not in this PR

- `.htaccess` HTTPS-redirect block activation — separate PR so existing shops upgrade independently.
- Security headers on runtime pages (non-installer) — separate PR.
- Strict `style-src` (no `'unsafe-inline'`) — separate follow-up; installer wizard still has legacy inline styles.

### DIFF
--- a/public_html/install/config
+++ b/public_html/install/config
@@ -11,12 +11,12 @@
 
 	define('FS_DIR_APP',         rtrim(str_replace('\\', '/', realpath(__DIR__.'/..')), '/') . '/');
 	define('FS_DIR_ADMIN',       FS_DIR_APP . 'backend/');
-	define('FS_DIR_STORAGE',     FS_DIR_APP . '{STORAGE_FOLDER}/');
+	define('FS_DIR_STORAGE',     FS_DIR_APP . '{STORAGE_FOLDER}' . '/');
 
 	// Web System
 	define('WS_DIR_APP',         preg_replace('#^'. preg_quote(DOCUMENT_ROOT, '#') .'#', '', FS_DIR_APP));
 	define('WS_DIR_ADMIN',       WS_DIR_APP . BACKEND_ALIAS . '/');
-	define('WS_DIR_STORAGE',     WS_DIR_APP . '{STORAGE_FOLDER}/');
+	define('WS_DIR_STORAGE',     WS_DIR_APP . '{STORAGE_FOLDER}' . '/');
 
 	######################################################################
 	## Database ##########################################################
--- a/public_html/install/includes/bootstrap.inc.php
+++ b/public_html/install/includes/bootstrap.inc.php
@@ -1,0 +1,68 @@
+<?php
+
+	/**
+	 * Shared bootstrap for installer and upgrader entry points.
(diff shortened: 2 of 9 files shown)


timint:
I welcome a lot of new ideas when it comes to the installer. It's is over 10 years old, so old that the glass border effect around has become trendy again. I have many times thought of refactoring and redoing it entirely. And as it has grown large over the years perhaps even make it standalone. Meaning instead of downloading LiteCart just download an installer that downloads and extracts LiteCart. Feel free to make it even more awesome. 🙂 As long as it has minimalistic thinking and CLI support.

PR #485, opened 2026-04-18
[v3] Harden session expiry and password reset flow (PROJ-23)

Wachhund:
Small one — three tiny auth gaps that add up to a decent attack chain in the customer flow.

## Summary
| Area | Change |
|---|---|
| Session expiry | `nod_session::load()` now rejects rows with `expires_at <= now()` and rotates the session id on miss; `save()` populates `expires_at` on INSERT too, so fresh sessions are not immediately treated as expired |
| Reset token | `ent_customer::set_password()` clears `password_reset_token` in the same update as the new hash and resyncs `$this->previous`; the reset branch in `account/reset_password` now calls `$customer->save()` with `sessions_expiry = now` to revoke parallel sessions |
| Enumeration oracle | The request branch collapses the three distinct error messages (unknown / inactive / normal) into a single uniform success notice; unknown or inactive accounts run a randomized `usleep(200–500ms)` dummy to keep response time in the same ballpark as the real send path |
| Edit strength check | `edit.inc.php` now runs `password_check_strength()` against `new_password` instead of the current password (one-character typo) |
| Testability | New `customer::is_session_expired($row)` extracted from `init()` so the revoke check is directly testable instead of reproduced in test logic |

## Why
The customer session and password-reset chain had four related gaps from an internal audit:
- expired session rows stayed authoritative because `load()` never checked `expires_at`
- `password_reset_token` was cleared in memory but not persisted after `set_password()`
- the forgot-password form returned three distinct error messages, so it worked as an email-verification oracle
- `account/edit` ran the strength check against the old password, so a strong current password let a weak new one through

All four hang off the customer-session-and-reset chain; fixing only some leaves the chain alive.

## Tests
New platform tests (`tests/*.php`, run via `php .git-hooks/pre-commit.d/platform_tests.php`):
| Test | Covers |
|---|---|
| `session_expiry.php` | Expired row rejected, new id rotated, `customer::check_login()` / `administrator::check_login()` stay not-true; `save()`→`load()` round-trip for fresh sessions |
| `reset_token_invalidation.php` | `password_reset_token` cleared in the same update as the new hash, and no longer resolvable |
| `reset_session_revoke.php` | `customer::is_session_expired()` flags older / missing `customer_security_timestamp`, accepts newer, passes through for accounts without `sessions_expiry` |
| `password_strength_new.php` | `edit.inc.php` checks `new_password` (static + behavioural assertion) |

Regression spot-checks: `customer.php`, `administrator.php`, `cart.php`, `func_password.php` all green.

## Notes
- The uniform-response message uses a new translation key `success_reset_password_email_sent_uniform` with an English fallback; it lazy-seeds into `lc_translations` on first render (existing v3 convention).
- The original "≤ 100 ms timing delta" target is not reachable with synchronous SMTP — the important part is that the message is uniform and the dummy path stays in the same order of magnitude as the real send. A truly constant-time response would need an async mail queue, which is post-3.0.
- I kept the preexisting `$_POST[$field]` bug in `edit.inc.php:76` out of this PR; it's in Wachhund:fix/edit-account-email-field → #483 as its own one-liner.
- Heads-up on CI: the Platform Tests job prints green but `continue-on-error: true` masks a preexisting `tests/database.php` failure that stops the runner before my new tests run. Wachhund:fix/platform-tests-database → #484 fixes that (one-liner).

## Test plan
- [ ] Apply #484 first (or run the suite locally) so the four new tests actually execute
- [ ] `php .git-hooks/pre-commit.d/platform_tests.php` — the four new tests report `[OK]`
- [ ] Manually: reset-password form for known / unknown / inactive account all return the same success notice
- [ ] Manually: account/edit with strong current + weak new password is rejected
- [ ] Manually: after a reset, a previously authenticated session on another browser is invalidated on the next request

### DIFF
--- a/public_html/frontend/pages/account/edit.inc.php
+++ b/public_html/frontend/pages/account/edit.inc.php
@@ -68,7 +68,7 @@
 					throw new Exception(t('error_passwords_missmatch', 'The passwords did not match.'));
 				}
 
-				if (!f::password_check_strength($_POST['password'])) {
+				if (!f::password_check_strength($_POST['new_password'])) {
 					throw new Exception(t('error_password_not_strong_enough', 'The password is not strong enough'));
 				}
 			}
--- a/public_html/frontend/pages/account/reset_password.inc.php
+++ b/public_html/frontend/pages/account/reset_password.inc.php
@@ -28,15 +28,13 @@
 				limit 1;"
 			)->fetch();
 
-			if (!$customer) {
-				throw new Exception(t('error_email_not_in_database', 'The email address does not exist in our database.'));
-			}
-
-			if (empty($customer['status'])) {
-				throw new Exception(t('error_account_inactive', 'Your account is inactive, contact customer support'));
-			}
-
 			if (!empty($_REQUEST['reset_token'])) {
+
+				// Reset-token branch: the token itself is the secret, so unknown or
+				// inactive accounts surface as a generic "invalid reset token" error.
+				if (!$customer || empty($customer['status'])) {
+					throw new Exception(t('error_invalid_reset_token', 'Invalid reset token'));
+				}
 
 				if (!$reset_token = json_decode($customer['password_reset_token'], true)) {
 					throw new Exception(t('error_invalid_reset_token', 'Invalid reset token'));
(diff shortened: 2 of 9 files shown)


timint:
You are doing amazing work!  👍It's great to have somone with a fresh pair of eyes see what needs to be done.

Issue #478, opened 2026-04-17
[v3] Admin auth hardening: remember-me 2FA, session revoke, enumeration


Wachhund:
## Symptom

Four admin-side defects:

- Remember-me cookie bypasses 2FA entirely — after HMAC verify, `nod_administrator.inc.php` loads the admin without setting `security_verification`. Stolen cookie = full admin session, even for admins with 2FA configured.
- Password / 2FA / permission changes don't revoke existing admin sessions. Three independent bugs: the edit form writes the timestamp under one key, `save()` doesn't persist it, and the enforce path in `nod_administrator::init()` reads a different key that was never written.
- Backend login distinguishes four failure modes (not found / disabled / blocked / expired) — admin enumeration oracle. No dummy `password_verify()` for timing, unlike the customer sign-in which was recently hardened.
- Remember-me cookies (admin and customer) set `Secure` only when `$_SERVER['HTTPS']` is set, not honoring `X-Forwarded-Proto` — unlike the session cookie in `nod_session`, which is proxy-aware.

## Direction

- Remember-me path runs the 2FA gate (sets `security_verification` if 2FA is configured, honors `known_ips`).
- Consolidate the session-revoke column name across `edit_administrator.inc.php`, `ent_administrator::save()`, and `nod_administrator` enforce path.
- Uniform error + dummy `password_verify()` in backend login.
- Shared `is_secure_request()` helper used by all three cookie call sites.

## Severity

HIGH (2FA bypass, stale admin sessions after credential rotation)

## References

Will be addressed by an upcoming PR (internal tracking: PROJ-24).

timint:
I saw your TOTP addon, it's very clean. Do you think we should add something to replace the current 2FA implementation entirely? I welcome all minimalistic upgrades that is mostly industry standard and doesn't require an entire 3rd party library. :)

By the look of all your powerful commits. I get the impression that you learned this project very quickly, and seem to find your way through files like it's your own backyard. 😄

Wachhund:
Appreciate the kind words! Three things, since you asked.

## 1. Replacing the current 2FA with TOTP

My read is that the two mechanisms solve slightly different problems, and replacing one-for-one would lose signal:

- The current email OTP (`backend/pages/verify.inc.php`) is an **unknown-IP challenge** — fires only when `$_SERVER['REMOTE_ADDR']` isn't in the admin's `known_ips`. It's a reactive speed-bump against credential stuffing / session replay from a new location, not a persistent 2nd factor.
[...]

PR #491, opened 2026-04-20
[v3] Fold TOTP 2FA add-on into administrator auth

Wachhund:
Follow-up to #478 — wiring up the TOTP proof-of-concept you asked about.

The existing unknown-IP email OTP stays exactly as-is. TOTP lives on top of it as an opt-in second factor: once an admin enrolls, every login (from any IP) goes through the TOTP challenge. Admins who don't enroll keep the current behaviour.

## Summary

| Area | Change | File |
|---|---|---|
| Helpers | New dependency-free TOTP module (base32 + `hash_hmac('sha1')`, inline-SVG QR generator, RFC 6238) | `public_html/includes/functions/func_totp.inc.php` (new) |
| Storage | Single new column `administrators.totp_secret VARCHAR(64) NULL` — no migration needed, picked up by the structure.json auto-patcher | `public_html/install/structure.json` |
| Login flow | After password verify, admins with `totp_secret` are redirected to the verify page with `security_verification.type = 'totp'` before the known-IP check runs | `public_html/backend/pages/login.inc.php` |
| Verify page | Accepts TOTP codes when `type = 'totp'`, otherwise keeps the existing email-OTP path. Known-IP trust is only updated via the email-OTP branch (TOTP is location-independent on purpose). Hides "Resend Code" in TOTP mode | `public_html/backend/pages/verify.inc.php` |
| Admin UI | New TOTP section on the edit-administrator page: Enable -> QR code + manual setup key + confirm code -> Disable (requires password re-entry) | `public_html/backend/apps/administrators/edit_administrator.inc.php` |

## Design choices

- **Layered, not replacing.** Email OTP is an unknown-IP challenge; TOTP is a permanent 2nd factor. Keeping both means admins without authenticator apps keep the IP-based speed-bump, enrolled admins get strong 2FA on every login. Matches GitHub / GitLab / Bitwarden.
- **No third-party library.** Base32 + HMAC-SHA1 are a few dozen lines; the QR generator (Version 1–6, ECC L, byte mode) is ~200 lines of straight PHP. No vendor tree, no Composer lift.
- **Recovery deliberately minimal.** No backup codes in v1 — if an admin loses their authenticator, another admin disables TOTP for them or sets `totp_secret = NULL` directly in the table. Can add backup codes later if desired; wanted to keep the first cut focused.
- **Session gate stays.** `nod_administrator::init()` already redirects to `verify` whenever `security_verification` is pending, so a half-logged-in admin can't touch other backend routes during the TOTP step.

## Test plan

- [x] `php -l` clean on all five touched files
- [x] Platform Tests job: all 60 tests green on this branch (including the four PROJ-23 tests and the existing admin/customer/cart suites)
- [ ] Manual: log in as an admin without TOTP — email OTP still triggers on unknown IP (unchanged behaviour)
- [ ] Manual: enroll TOTP on the edit-administrator page, scan QR with Google Authenticator / Authy / 1Password, confirm with the 6-digit code — `totp_secret` populated
- [ ] Manual: log in again — TOTP challenge replaces the email flow, login from a known IP still triggers the TOTP step
- [ ] Manual: disable TOTP with correct password — `totp_secret` cleared; wrong password rejected
- [ ] Manual: half-way through enrollment (pending secret in session), reload the admin edit page — the pending confirm UI persists until confirmed or the page is left alone across a session reset

## Not in scope here

- Backup/recovery codes (can add later if you want)
- Customer-side TOTP (PROJ-11 in my notes, separate PR)
- Remember-me 2FA integration (one of the other items in #478)
- Translation strings beyond the fallback English — lazy-seeded per the v3 convention; a dedicated seed pass can follow

Marked as a proof-of-concept per the "happy to spec that" offer in #478. Pick it apart freely — happy to rework the storage shape, split enrollment into a dedicated app, or change the UI placement if you'd prefer.

### DIFF
--- a/public_html/backend/apps/administrators/edit_administrator.inc.php
+++ b/public_html/backend/apps/administrators/edit_administrator.inc.php
@@ -15,6 +15,68 @@
 	breadcrumbs::add(t('title_administrators', 'Administrators'), document::href_ilink(__APP__.'/administrators'));
 	breadcrumbs::add(!empty($administrator->data['username']) ? t('title_edit_administrator', 'Edit Administrator') : t('title_create_new_administrator', 'Create New Administrator'));
 
+	// TOTP enroll/confirm/disable — handled before the main save so the sub-form
+	// buttons (totp_setup, totp_confirm, totp_disable) don't have to go through
+	// the generic save validation.
+	if (!empty($administrator->data['id']) && (!empty($_POST['totp_setup']) || !empty($_POST['totp_confirm']) || !empty($_POST['totp_disable']))) {
+
+		require_once 'app://includes/functions/func_totp.inc.php';
+
+		try {
+
+			if (!empty($_POST['totp_setup'])) {
+				session::$data['totp_pending_secret'] = totp_generate_secret();
+				redirect(document::ilink());
+				exit;
+			}
+
+			if (!empty($_POST['totp_confirm'])) {
+
+				if (empty(session::$data['totp_pending_secret'])) {
+					throw new Exception(t('error_totp_setup_expired', 'TOTP setup session expired. Please try again.'));
+				}
+
+				if (empty($_POST['totp_code']) || !totp_verify_code(session::$data['totp_pending_secret'], $_POST['totp_code'])) {
+					throw new Exception(t('error_invalid_verification_code', 'Invalid verification code'));
+				}
+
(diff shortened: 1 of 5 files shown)


timint:
Absolutely high class standard man! 👍🙂

Issue #492, opened 2026-04-21
error in edit_product.inc.php when searching category where user has app restrictions


averlon:
### Description

pleas have a look at issue #469. It looks like as if it is a bug!

### LiteCart Version

2.6.3

### PHP Version

8.3

### Error/Backtrace

```shell

```

### Is this a problem that can be reproduced in the demo platform (https://demo.litecart.net/admin)?

None

### If this a visual problem that could be related to a browser. Which one?

None

Wachhund:
@averlon — excellent detective work, your diagnosis is correct. This is a real bug in vanilla 2.6.3 (and the same logic exists in v3's `dev-major`, so it'll need to be fixed there too).

## Root cause

The permission gate in `admin/index.php` around line 97:

```php
if (empty(user::$data['apps']) || (!empty(user::$data['apps'][$_GET['app']]['status']) && in_array($_GET['doc'], user::$data['apps'][$_GET['app']]['docs']))) {
```

Treats every doc the same, including **helper endpoints** that aren't shown in the admin's permission UI — things like `categories.json`, `attribute_values.json`, `category_picker`, `product_picker`, etc. Those are declared in each app's `docs` list (see `admin/catalog.app/config.inc.php` → `'categories.json' => 'categories.json.inc.php'`) and are dispatched through the same `?app=…&doc=…` router as the visible views, but the administrator-edit page only offers you the `menu` items to tick. So a restricted user ends up with `docs = ['catalog', 'edit_product']`, opens the product editor, the category picker fires an AJAX call to `categories.json`, and the `in_array()` check rejects it → `Access Denied` → the picker receives HTML instead of JSON → the `SyntaxError: Unexpected token '<'` on the client side.

Your manual workaround (adding `categories.json` to the docs array directly in the DB) works, but doesn't scale — every `.json` helper in every app would need the same treatment, and a fresh install of a new helper would hit the same wall.

## Proposed fix (sketch)

Recognise helper docs implicitly when the admin has access to the app at all:

```php
// v3 — backend/pages/index.inc.php around line 22
$is_helper = preg_match('/\.(json|csv)$/', __DOC__)
          || str_ends_with(__DOC__, '_picker');

if (!empty(administrator::$data['apps'][__APP__]['status'])
    && !$is_helper
    && !in_array(__DOC__, administrator::$data['apps'][__APP__]['docs'])) {
    notices::add('errors', t('title_access_denied', 'Access Denied'));
    return;
}
```

[...]

Issue #469, opened 2026-04-12
no bug - but help requested for edit_product.inc.php


averlon:
### What where you trying to do?
I am looking through the code of litecart 2.6.3.

When editing a product, in the information tab there is the list of categories and below this list there is a search field.

I would need some hints where/how this search filed is processed. I wasn't able to identify the code for this!

### LiteCart Version

2.6.3

### PHP Version

8.3

### Error/Backtrace

```shell

```

### Is this a problem that can be reproduced in the demo platform (https://demo.litecart.net/admin)?

No

### If this a visual problem that could be related to a browser. Which one?

Not relevant

Wachhund:
The category search field on the product edit page is powered by a few components working together:

**1. HTML rendering** — `includes/functions/func_form.inc.php` line 983, function `form_draw_categories_list()`. This builds the category list, the search input, and initializes the JS widget.

**2. JS widget** — jQuery plugin `categoryPicker` (look for it in the admin template JS). It's initialized at line 1030 of `func_form.inc.php` and binds to elements with `data-toggle="category-picker"`.

**3. Data endpoint** — The picker fetches categories from a JSON endpoint: `admin/?app=catalog&doc=categories.json`. This is the file `admin/catalog.app/categories.json.inc.php` which returns the category tree as JSON.

**4. Search logic** — The search input triggers an AJAX call to the JSON endpoint. The filtering/matching happens client-side in the `categoryPicker` plugin against the returned category names.

So the flow is:
```
Search input → categoryPicker JS → AJAX to categories.json → filter results → display dropdown
```

The key files to look at:
- `includes/functions/func_form.inc.php` (line ~983)
- `admin/catalog.app/categories.json.inc.php`
- Admin template JS (search for `categoryPicker`)

Hope that helps!

averlon:
@Wachhund thanks for these hints!

I can see the search-field in the view.
When I put the cursor into the search-field en empty field opens below the search-field.
When entering some search-string I get an error ("not found").

I currently don't have a clou why the error message shows up!

Wachhund:
Found it — the endpoint unintentionally AND's your search term with the current `parent_id`, so it only matches categories directly under the parent you're browsing.

In `admin/catalog.app/categories.json.inc.php`:
- line 5: `$_GET['parent_id']` defaults to `0`
- line 32: the query always appends `and c.parent_id = <parent>` (because `isset()` is still true after that default)
- line 33: adds the search filter on top

[...]

PR #494, opened 2026-04-21
[v3] Fix restricted admins getting Access Denied on JSON/picker helpers

Wachhund:
Same bug the v2 PR addresses, just in the v3 router.

`backend/pages/index.inc.php` gates each admin doc by strict allow-list match. The administrator edit UI only lets you tick the visible menu docs, so auxiliary endpoints declared in each app's `config.inc.php` — `categories.json`, `attribute_values.json`, `category_picker`, `product_picker`, etc. — are never in a restricted admin's docs array. Restricted admins open the product editor, the picker fires AJAX, and the gate rejects the helper with Access Denied.

## Summary
| Change | File |
|---|---|
| Allow helper docs (`*.json`, `*.csv`, `*_picker`) implicitly when the administrator has the app enabled | `public_html/backend/pages/index.inc.php` |

## Why
- Unrestricted admins unaffected (status check already bypasses them).
- Restricted admins without the app enabled are still refused — helper-status is not a bypass for app-level disablement.
- Pattern covers every helper I can find in the v3 `backend/apps/*/config.inc.php` files.

## Test plan
- [ ] Create a restricted administrator with `apps = {"catalog":{"status":"1","docs":["catalog","edit_product"]}}`
- [ ] Log in and open a product → picker search hits `categories.json` and returns JSON
- [ ] Unknown helper (`?app=catalog&doc=nonexistent.json`) — still rejected by the app's doc resolver (line 28)

Reported upstream on v2 by @averlon in litecart/litecart#469 and litecart/litecart#492; fix is the same on both branches. The v2 PR is tracked separately.

### DIFF
--- a/public_html/backend/pages/index.inc.php
+++ b/public_html/backend/pages/index.inc.php
@@ -18,8 +18,16 @@
 			'color' => $app_config['theme']['color'] ?? '#97a3b5',
 		];
 
-		// Check if administrator is permitted to access document
-		if (!empty(administrator::$data['apps'][__APP__]['status']) && !in_array(__DOC__, administrator::$data['apps'][__APP__]['docs'])) {
+		// Check if administrator is permitted to access document.
+		// Helper endpoints (*.json, *.csv, *_picker) aren't individually tickable in the
+		// administrator edit UI — they're auxiliary routes consumed by the main docs.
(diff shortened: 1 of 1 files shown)


timint:
Supergreat!

PR #499, opened 2026-04-28
[v2] Fix Geo Zones pending Add row dropped silently on Save (#498)

### LINKED ISSUE #498: Geo Zones: pending row in Add form is silently dropped on Save unless 'Add' is clicked first
### What happens

When editing a geo zone in admin (`Geo Zones → Edit/Create`), the user fills the Country/Zone/City fields in the bottom "Add" row and clicks **Save** at the top of the page. The new row is **silently discarded** — the geo zone is saved with zero zone rows.

### Workaround that the user is expected to figure out

Click the **Add** button in the row first, *then* Save. Reported by @jay2216 in the LiteCart forum — the original symptom was "Zone Based Shipping always picks the Non-matched rate table". Root cause turned out to be a saved-but-empty geo zone caused by exactly this UX trap.

### Root cause

`public_html/admin/geo_zones.app/edit_geo_zone.inc.php`

| Step | Code | Behavior |
|------|------|----------|
| Form input row (line 119–123) | `name="new_zone[country_code]"`, `name="new_zone[zone_code]"`, `name="new_zone[city]"` | Lives in `<tfoot>`. Singular `new_zone` namespace. |
| JS Add handler (line 174–213) | Clones values into `<tbody>` as `zones[new_X][...]` | Only fires on `<button name="add">` click. |
| Save handler (line 18–44) | Reads `$_POST['zones']` only | `$_POST['new_zone']` is **never consulted**. |

So if the user submits without clicking Add, the entire `new_zone` payload is thrown away server-side.

### Suggested fix (server-side, JS-independent)

In the save block, fold a non-empty `new_zone` into `zones` before persisting:

```php
if (isset($_POST['save'])) {

  try {

    if (empty($_POST['zones'])) $_POST['zones'] = [];

  // Pick up a pending row that the user filled but didn't add
    if (!empty($_POST['new_zone']['country_code'])) {
      $_POST['zones']['new_pending'] = $_POST['new_zone'];
    }

    $fields = ['code', 'name', 'description', 'zones'];

    foreach ($fields as $field) {
      if (isset($_POST[$field])) $geo_zone->data[$field] = $_POST[$field];
    }

    $geo_zone->save();
```

Server-side absorption is preferable to a JS-only fix because:

- Works when the user submits via Enter (default form submit, no Add click).
- Works when JS is disabled or fails to load.
- Mirrors the behavior an unsuspecting user already expects.

### Related (v3)

In the forum reply Tim mentioned that v3 already has an alert reminder when the Add button wasn't pressed. This issue is to bring the same robustness to v2 — server-side absorption is arguably the cleaner approach since it skips the confirmation dialog entirely.

### Repro (≈5 sec, also reproducible on the demo)

1. Admin → Geo Zones → New
2. Code/Name = `cr` / `Costa Rica`
3. In the bottom row pick Country = Costa Rica, leave Zone/City blank
4. Click **Save** at the top
5. Reopen the geo zone — zone list is empty

### Test plan

- [ ] Edit an existing geo zone, fill the new-row Country/Zone, hit Save (no Add click) → row persists.
- [ ] Same but pressing Enter inside the City field → row persists.
- [ ] Filling only Country (no Zone/City) → row still persists with "All Zones / All Cities" semantics.
- [ ] Leaving the Add row completely empty + Save → no spurious empty row created.
- [ ] Existing rows + new row simultaneously → all persist.
- [ ] After save, "Zone Based Shipping" module evaluates the geo zone correctly (no fallthrough to "Non-matched Zones").

Wachhund:
Quick follow-up to #498 — pending row absorbed server-side now, JS path unchanged.

## What

`new_zone[...]` is the input row in the `<tfoot>`. The JS Add handler clones it into `<tbody>` as `zones[new_X][...]`, but if the user submits without clicking Add, the save handler only reads `$_POST['zones']` and the entire `new_zone` payload is silently dropped.

Two-line server-side fix: if `new_zone` has a country code, fold it into `zones` before persisting.

## Why server-side

| | |
|---|---|
| Default Enter submit | Pressing Enter inside the City field doesn't trigger the Add click |
| JS disabled / failed load | Form still submits, row would still be lost |
| Mirrors user intent | What the user already expected to happen |

## Test plan

- [ ] Edit geo zone, fill new-row Country/Zone, hit Save (no Add click) → row persists
- [ ] Press Enter inside the City field → row persists
- [ ] Country only (Zone/City blank) → row persists with All Zones / All Cities semantics
- [ ] Empty Add row + Save → no spurious row created
- [ ] Existing rows + new row simultaneously → all persist
- [ ] Zone Based Shipping evaluates the zone correctly post-save (no fallthrough to Non-matched)

Closes #498

### DIFF
--- a/public_html/admin/geo_zones.app/edit_geo_zone.inc.php
+++ b/public_html/admin/geo_zones.app/edit_geo_zone.inc.php
@@ -21,6 +21,11 @@
 
       if (empty($_POST['zones'])) $_POST['zones'] = [];
 
+    // Fold a pending row that was filled but not added via the Add button
+      if (!empty($_POST['new_zone']['country_code'])) {
+        $_POST['zones']['new_pending'] = $_POST['new_zone'];
+      }
+
       $fields = [
         'code',
         'name',


timint:
Wunderbar, the most lightweight solution I can think of :)

PR #509, opened 2026-05-01
[V2.6.3] e-Mail confirmation for new customers

averlon:
I was able to create an e-Mail confirmation for new customers.

When a new customer registers the e-Mail address was not validated/confirmed before the customer account was activated.

In my setup, the customer account is deactivated by default after registration and an e-Mail confirmation message is sent to the e-Mail address given in the registration process. If the link is touched, the account is activated.


averlon:
I was able to create an e-Mail confirmation for new customers.

averlon:
the naming-conventions for "av_account_confirmation" in public_html\includes\templates\default.catalog\pages\account_confirmation.inc.php may not fit!

For DE I defined it in some csv-file:

av_account_confirmation,Ihre e-Mail Adresse wurde bestätigt. Ihr Kundenkonto ist jetzt aktiviert! Sie können die Forstinninge Güterbörse jetzt nutzen.

For EN probably: Your e-Mail address has been confirmed. You account has been activated!

timint:
Sorry this feature requires some security added to the urls or anyone can access them.
For 3.0 I can add the same type of verifications as we have on litecart.net with a 6-digit code. I think I would like also add a setting for verifications.

Account Verifications
[x] Disabled
[ ] When creating an account only
[ ] When signing in with a previously not known IP address

Does that sound like a nice feature to have?

averlon:
@timint your project -> your decision!

1) if I am not wrong, 2.6.3 didn' t have any confirmation for the new customer account. So my implementation is probably a step forward - at least for my implementation based on 2.6.3.
2) I have several accounts in my password manager. Forums or Shops and more. As far as I remember, none of them had a different approach to confirm the e-Mail of the new account is valid. Man-in-the-middle attacks are very rare. And, based on my knowledge, these guys have other options to threaten you if thy are on you! At the point of the confirmation process there is nothing stored in the account as address, e-Mail and phone number. No bank account, no credit card, nothing of value for a hacker (if I am not wrong). So the process is based on the asumption, that only the one requesting the account has access to the e-Mail address provided for the account. Some assumptions should show true in our IT-World, else we can shutdown most of our actions.
3) Your proposal: 1) yes, 2) yes, 3) no - why? (this is a rhethorical question). I can only talk for DE. We have a very strict law (DSGVO) what you are allowed to store and what not. I have seen several fields in the DB like IP-address, Date loged-in .... I doubt they complie with the DSGVO. At least you would need a side long text on Privacy Policy or Data Security Policy and Cookies and even then it is doubtful it will be within the law (my comment: we are crazy on that!).

As I already stated, my knowledge on PHP is very limited. For more steps on security I am for sure not able to help or provide code snippets. I will later then have a look how you did it last but not least on V3 once it will be released!

timint:
1. Yes, but the confirmation url contains no security so an attacker can confirm any email address they desire without even opening a single email.
>` document::ilink('account_confirmation', ['customer_id' => $customer->data['id'], 'customer_email' => $customer->data['email']], false, [], $language_code)`

https://mysite.tld/account_confirmation?

I'm not saying email verifications is a bad idea. I'm saying doing that with urls that have no security or uniqueness is a bad idea. I provided a solution to that in my reply.

2. This is not a man in the middle problem. This is "any man" can change the email address in the url and confirm any email address. It is not protected with any unique code bound to the email. The whole point with email verifications is that you are required to open the email.

Prove me I'm wrong. I feel I'm taking a hits here for pointing out a problem with this pull request.

3. I have no knowledge of DSGVO and maybe that could be a separate topic. I understand it's GDPR and the ePrivacy Directive on steroids. LiteCart does provide a way to publish a privacy policy and connect it to consent collecting. But some things are inevitable. You can't access a hostname without leaving a trace in the logs. That trace will contain IP and a timestamp. That's every server on the web. There are even laws that enforce traffic logging. Now what's critical is sharing that with third parties like Google Analytics or saving it for an unreasonable amount of time. Keeping it to yourself for a fair period of time is usually not an offense. Again, separate topic.

EDIT: I see that you added customer_id to the url which function as some sort of uniqueness. But that is too poor for security as it's a incremental number and can be used publicly in other parts of the platform. It's not safe to substitute a shared secret confirmation code.

averlon:
@timint 
`This is "any man" can change the email address in the url and confirm any email address.`

Yes, you are right. I didn't recognize this!

I tried to copy the code for the "printable copy of an order". I think, this is unique. I will look in more detail into it and try to modify the aproach (as far as I can)!

Thanks for clarification!

timint:
What I will do is add email verifications to 3.0, similar to what we have on litecart.net. I will make it an optional feature to disable.

The order copy has public_key which provides uniqueness to the url and is a lengthy process (annoying) to brute force.

averlon:
I managed now to make the link like this:

`http://s42252d6.averlon.loc:8080/de/account_confirmation?public_key=vum5pruy3g8ydknlx2j3lvuptqm28qp2`

timint:
You might want to look into what SQL injections are. This just gave the attacker full access to your database.

`      where public_key = '". $_GET['public_key'] ."'`

PR #511, opened 2026-05-03
[v3] Fix administrator known_ips/known_fingerprints array drift (#503)

### LINKED ISSUE #503: [v3] Platform test administrator.php fails: administrator::$data does not match $administrator->data
Surfaced after #502 unblocked the schema import — `Run platform tests` now reaches the runner and the very first test (`administrator.php`) fails on the static-node vs entity comparison.

## Symptom

```
Found 65 test files
Running tests from administrator.php... [Failed]
Error: administrator::$data does not match $administrator->data [FAIL]
##[error]Process completed with exit code 1.
```

The runner aborts on the first failure, so the remaining 64 tests don't get to run. Job stays green only because of `continue-on-error: true` in `.github/workflows/ci.yml:282`.

## Root cause sketch

`tests/administrator.php:109-113`:

```php
administrator::load($administrator->data['id']);

if (!f::array_intersect_compare($administrator->data, administrator::$data)) {
  throw new Exception('administrator::$data does not match $administrator->data');
}
```

The test compares the entity layer (`ent_administrator->data`) with the static node layer (`nod_administrator::$data` exposed as `administrator::$data`). After saving + reloading, the two diverge.

`nod_administrator` has seen a series of recent changes that the test was not updated to match:

| Commit | Touched |
|--------|---------|
| `9c3604446` | Replaced SHA1 remember-me with HMAC-SHA256 tokens |
| `adbca63be` | Mixed minor adjustments |
| `6058368a7` | Mixed fixes and polishing |
| `488f2f2d2` | Minor adjustments |
| `668c34054` | Added known-fingerprint tracking, increased known IPs to 10 |

Suspect candidates for the drift:

- New / renamed columns surfacing on `administrator::$data` after `load()` but not present on the entity post-`save()` (or vice versa).
- `known_ips` / `known_fingerprints` JSON normalization differs between the static-node load path and the entity save path.
- Recently added `password_reset_*` / session fields on one side but not the other.

A quick before/after dump of both arrays in `tests/administrator.php` right before the assertion would point at the offending field.

## Repro

1. Schema imports clean (PR #502 lands first, otherwise CI crashes earlier).
2. `php .git-hooks/pre-commit.d/platform_tests.php` from repo root with the test DB seeded.
3. First test reports `[FAIL]` immediately.

Visible in CI run https://github.com/litecart/litecart/actions/runs/25085577253/job/73500343073 (PR #502).

## Severity

Low for now — `continue-on-error: true` keeps the build green, no functional regression in the admin UI. Worth fixing before the 3.0 release because it masks the other 64 tests that don't get a chance to run.

## Suggested next step

Either align the test with the new `nod_administrator` shape (drop / add the fields that diverge), or normalize the data shape between `ent_administrator::data` and `nod_administrator::$data` so they round-trip cleanly. Happy to take a stab if you want, but flagging here first since you've been the one driving the auth-layer changes.

Wachhund:
Following your hint in #503 — `f::string_split` on load. Plus a small typo fix in the edit form that I noticed on the way through.

## What

`tests/administrator.php` was failing at the `administrator::$data` vs `$administrator->data` assertion because the two layers stored `known_ips` / `known_fingerprints` in different shapes:

| Layer | Shape after load |
|-------|------------------|
| `nod_administrator::load()` (line 157-158) | array (split on `,`) |
| `ent_administrator::load()` | comma-separated string (raw DB value) |

`f::array_intersect_compare` saw `'1.2.3.4,5.6.7.8'` against `['1.2.3.4', '5.6.7.8']` and threw.

## Fix

| File | Change |
|------|--------|
| `includes/entities/ent_administrator.inc.php` | `load()` now applies `f::string_split()` to `known_ips` and `known_fingerprints`, matching the node. `reset()` initialises both to `[]` for the new-administrator path. |
| `backend/apps/administrators/edit_administrator.inc.php` | Switched the two readonly displays from `str_replace(',', ', ', $string)` to `implode(', ', $array)`, since the entity now hands the form an array. Also fixed a pre-existing typo: `$administrator->data['fingerprints']` → `$administrator->data['known_fingerprints']` (the form was reading a non-existent key). |

No save-side change needed — `ent_administrator::save()` doesn't write `known_ips` / `known_fingerprints`. Those columns are populated by direct UPDATE statements in `nod_administrator::init()` (login tracking) and `pages/verify.inc.php` (known-IP enrolment) using comma-string concat. Both keep working unchanged.

## Verified

- `php -l` clean on both files
- `f::string_split('')` returns `[]`, so the new-record / never-logged-in case is fine
- The `(string)` cast inside `string_split()` makes the call PHP 8.5 deprecation-safe even if the column is null

## Test plan

- [ ] `php tests/administrator.php` exits 0 (was: `[FAIL]` on first assertion)
- [ ] Edit administrator page renders without warnings — IP/fingerprint readonly blocks show a comma-space-separated list
- [ ] New administrator (id=null) form renders cleanly with empty IP/fingerprint blocks
- [ ] Existing admins continue to log in; `init()` and `verify.inc.php` write paths untouched

Closes #503

### DIFF
--- a/public_html/backend/apps/administrators/edit_administrator.inc.php
+++ b/public_html/backend/apps/administrators/edit_administrator.inc.php
@@ -346,14 +346,14 @@
 					<label class="form-group">
 						<div class="form-label"><?php echo t('title_known_ip_addresses', 'Known IP Addresses'); ?></div>
 						<div class="form-input" readonly style="height: 80px;">
-							<?php echo str_replace(',', ', ', $administrator->data['known_ips']); ?>
+							<?php echo implode(', ', $administrator->data['known_ips']); ?>
 						</div>
 					</label>
 
 					<label class="form-group">
 						<div class="form-label"><?php echo t('title_known_fingerprints', 'Known Fingerprints'); ?></div>
 						<div class="form-input" readonly style="height: 80px;">
-							<?php echo str_replace(',', ', ', $administrator->data['fingerprints']); ?>
+							<?php echo implode(', ', $administrator->data['known_fingerprints']); ?>
 						</div>
 					</label>
 					<?php } ?>
--- a/public_html/includes/entities/ent_administrator.inc.php
+++ b/public_html/includes/entities/ent_administrator.inc.php
@@ -25,6 +25,8 @@
 
 			$this->data['apps'] = [];
 			$this->data['widgets'] = [];
+			$this->data['known_ips'] = [];
+			$this->data['known_fingerprints'] = [];
 
 			$this->previous = $this->data;
 		}
@@ -53,6 +55,8 @@
 
 			$this->data['apps'] = !empty($this->data['apps']) ? json_decode($this->data['apps'], true) : [];
 			$this->data['widgets'] = !empty($this->data['widgets']) ? json_decode($this->data['widgets'], true) : [];
+			$this->data['known_ips'] = f::string_split($this->data['known_ips']);
+			$this->data['known_fingerprints'] = f::string_split($this->data['known_fingerprints']);
 
 			$this->previous = $this->data;
 		}


timint:
Just as I would have done it, literally. You are a perfect fit for this project!

PR #512, opened 2026-05-05
[v3] Run CI against MariaDB 12 and MySQL 8.4 LTS in matrix

### LINKED ISSUE #488: [v3] Run CI against MySQL in addition to MariaDB
Noticed the CI now targets MariaDB 12 (490653f7b) — which is great for the new testing baseline, but the stack profile still says "MySQL and MariaDB compatible" and there's no automated check keeping that promise. Thought it'd be worth floating the idea of a matrix build before we drift.

## What would change

Turn the `platform-tests` and `e2e-tests` jobs into a matrix that spins up both databases and runs the suite against each. Something like:

```yaml
strategy:
  fail-fast: false
  matrix:
    db:
      - { image: "mariadb:12", root_env: "MARIADB_ROOT_PASSWORD", user_env: "MARIADB_USER", health: "healthcheck.sh --connect --innodb_initialized" }
      - { image: "mysql:8.4",  root_env: "MYSQL_ROOT_PASSWORD",   user_env: "MYSQL_USER",   health: "mysqladmin ping -h localhost" }
```

The `SET SESSION sql_mode = ''` trick in the schema importer and the timezone handling in `+ Set mysql connection timezone` would both get exercised against the real MySQL side-by-side.

## Pros

| | |
|---|---|
| **Catch drift early** | Incompatibility shows up in PRs, not in an operator's inbox six months post-release |
| **Honour the stack promise** | The profile claims compatibility; matrix makes it enforceable |
| **Surface `sql_mode` gaps** | `chk_userdata` and similar CHECK-constraints behave differently under strict MySQL — matrix forces us to decide: fix the test or fix the constraint |
| **Cheap insurance for `json_value()`, timezones, fulltext** | The places where the two diverge historically |

## Cons

| | |
|---|---|
| **Longer CI wall time** | Platform-tests + E2E roughly double in duration (currently ~2–3 min each, so ~5–6 min added). Matrix runs in parallel, so wall-clock doubles less than total compute |
| **Flakier signal initially** | First run will almost certainly expose existing MySQL-side issues, giving a temporarily red column until we chase them down |
| **More cache lines** | Playwright + Composer caches stay single-keyed, but DB warmup is per-matrix-row |

## Versions I'd propose

- **MariaDB 12** (current) — already the baseline
- **MySQL 8.4 LTS** — the supported LTS release, closest to what most shared-hosting providers ship today

We could add **MySQL 9** later if there's appetite, but 8.4 covers the "does this actually work for our user base" question.

## Not in scope here

- Matrix over PHP versions (8.3 ↔ 8.5) — separate concern, leave as-is for now
- MariaDB 11 LTS backstop — only needed if we care about older-LTS deployments

## Question

Would you be open to a PR that wires this up? Happy to do the legwork on the `sql_mode` / timezone / CHECK-constraint gotchas that surface once MySQL joins the run. If you'd rather keep CI single-DB for now and track compatibility another way, that works too — just want to make sure we don't quietly lose the promise.

Wachhund:
Hey, picking up #488 — turning `platform-tests` and `e2e-tests` into a 2×DB matrix so the *MySQL and MariaDB compatible* line in `STACK_PROFILE` is something CI actually proves on every PR.

First matrix run did exactly what it was supposed to: surfaced a real MySQL-8.0+ incompatibility in the schema importer that MariaDB had been swallowing quietly. Fixed it in the same PR rather than file a follow-up — see commit 2.

## Commits

| | |
|---|---|
| `+ Run CI against MariaDB 12 and MySQL 8.4 LTS in matrix` | The matrix itself — `fail-fast: false`, parametrised service container, per-engine artefact names |
| `+ Wrap DEFAULT in parens for BLOB/TEXT/JSON/GEOMETRY columns` | The compat fix the matrix surfaced. MySQL 8.0+ rejects bare `DEFAULT` literals on these column types; both engines accept `DEFAULT (…)` in parens |

## Matrix details

| | |
|---|---|
| Strategy | `fail-fast: false` over `mariadb:12` + `mysql:8.4` for both `platform-tests` and `e2e-tests` |
| Service env | `MYSQL_*` only — MariaDB 10.6+ accepts those as legacy; no per-engine env switching |
| Health-check | parametrised: `healthcheck.sh` for MariaDB, `mysqladmin ping` for MySQL |
| Health-retries | `5 → 10` for MySQL's slightly longer first boot |
| Artefacts | `playwright-report-${{ matrix.db.slug }}` so the matrix rows don't clobber each other |
| Drift fix | `e2e-tests` was on `mariadb:11` while `platform-tests` was on `mariadb:12` — both now on 12 |

## Why the DEFAULT-paren fix touches three places

The schema-build logic is duplicated three times: `install.php` (initial install), `upgrade.php` (schema upgrades on existing installs), and the CI importer in `ci.yml` (two matrix copies). All three needed the same patch. Worth a follow-up to consolidate them, but out of scope here.

## Out of scope

| | |
|---|---|
| PHP version matrix (8.3 ↔ 8.5) | Separate concern |
| MariaDB 11 LTS / MySQL 9 | LTS 8.4 covers shared-hosting; older/newer can be added later |
| Composite action for the importer | Matrix doesn't multiply YAML duplication — block stays at two copies |
| Consolidating install.php / upgrade.php / CI schema-build into one source | Cleaner, bigger blast radius |

## Heads-up after merge

Branch protection on `dev-major` likely lists `Platform Tests` and `E2E Tests (Playwright)` as required status checks. After merge those names disappear — you'll want to update required checks to the four new matrix-suffixed names (`Platform Tests (mariadb-12)`, `Platform Tests (mysql-8.4)`, `E2E Tests (mariadb-12)`, `E2E Tests (mysql-8.4)`).

## Test plan

- [x] Lint + Build green
- [x] `Platform Tests (mariadb-12)` green
- [x] `Platform Tests (mysql-8.4)` green
- [x] `E2E Tests (mariadb-12)` green
- [x] `E2E Tests (mysql-8.4)` green
- [ ] Branch protection on `dev-major` updated to require the four new matrix-suffixed checks (post-merge, on your side)

Closes #488

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -242,7 +242,8 @@ jobs:
                 if (!empty(\$def['length'])) \$type .= '(' . \$def['length'] . ')';
                 if (!empty(\$def['unsigned'])) \$type .= ' UNSIGNED';
                 \$null = (!isset(\$def['null']) || \$def['null']) ? '' : ' NOT NULL';
-                \$default = isset(\$def['default']) ? ' DEFAULT ' . \$def['default'] : '';
+                \$blob_text = in_array(strtoupper(\$def['type']), ['TEXT','TINYTEXT','MEDIUMTEXT','LONGTEXT','BLOB','TINYBLOB','MEDIUMBLOB','LONGBLOB','JSON','GEOMETRY']);
+                \$default = isset(\$def['default']) ? ' DEFAULT ' . (\$blob_text ? '(' . \$def['default'] . ')' : \$def['default']) : '';
                 \$auto = !empty(\$def['auto_increment']) ? ' AUTO_INCREMENT' : '';
                 \$cols[] = \"\`\$col\` \$type\$null\$default\$auto\";
               }
@@ -390,7 +391,8 @@ jobs:
                 if (!empty(\$def['length'])) \$type .= '(' . \$def['length'] . ')';
                 if (!empty(\$def['unsigned'])) \$type .= ' UNSIGNED';
                 \$null = (!isset(\$def['null']) || \$def['null']) ? '' : ' NOT NULL';
-                \$default = isset(\$def['default']) ? ' DEFAULT ' . \$def['default'] : '';
+                \$blob_text = in_array(strtoupper(\$def['type']), ['TEXT','TINYTEXT','MEDIUMTEXT','LONGTEXT','BLOB','TINYBLOB','MEDIUMBLOB','LONGBLOB','JSON','GEOMETRY']);
+                \$default = isset(\$def['default']) ? ' DEFAULT ' . (\$blob_text ? '(' . \$def['default'] . ')' : \$def['default']) : '';
                 \$auto = !empty(\$def['auto_increment']) ? ' AUTO_INCREMENT' : '';
                 \$cols[] = \"\`\$col\` \$type\$null\$default\$auto\";
               }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -151,25 +151,36 @@ jobs:
           npx gulp scss-trumbowyg
 
   platform-tests:
-    name: Platform Tests
+    name: Platform Tests (${{ matrix.db.slug }})
     runs-on: ubuntu-latest
     needs: [lint]
 
+    strategy:
+      fail-fast: false
+      matrix:
+        db:
+          - image: mariadb:12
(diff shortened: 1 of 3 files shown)


timint:
I see this broke with my latest commits to move data.sql into data/*.json. Let me adapt this to some CI adjustments I am making. So I will patch this in a future commit.

PR #517, opened 2026-05-13
Update for PHP 8.0

s22-tech:


Wachhund:
Nice modernization — `str_contains()` reads cleaner than the `=== false` dance.

Two small notes:

- There's one more eligible spot in runtime code that could move along: `public_html/includes/nodes/nod_route.inc.php:75`. (The two in `install/install.php` and `install/upgrade.php` are bootstrap-phase, so probably fine to leave for now.)
- The extra space before the `//` comments looks like an unintended whitespace tweak — the rest of v3 sticks to a single space, so it might be worth reverting that part to keep the diff focused on the PHP 8 change.

Otherwise behavior is 1:1 and the swap is safe. Thanks!

s22-tech:
> There's one more eligible spot in runtime code that could move along: public_html/includes/nodes/nod_route.inc.php:75. (The two in install/install.php and install/upgrade.phpare bootstrap-phase, so probably fine to leave for now.)
	I plan on doing the others, as well, now that 8.0 is the base.  Since I don't do PR's all that often, I like to start off slow in case I make a mistake.

> The extra space before the // comments looks like an unintended whitespace tweak — the rest of v3 sticks to a single space, so it might be worth reverting that part to keep the diff focused on the PHP 8 change.
	I reverted those on my branch.  Hopefully they're viewable upstream.

Wachhund:
Thanks for expanding it — nice sweep across the codebase.

Spotted one bug worth catching before merge:

`public_html/includes/entities/ent_image.inc.php` line 97 has a typo — `srt_contains` instead of `str_contains`:

```php
if (srt_contains(file_get_contents($this->_file, false, null, 0, 256), '<svg')) {
```

That path runs when `exif_imagetype()` returns `false` (i.e. the SVG-detection branch), so it'd throw a fatal as soon as someone uploads an SVG.

One small style note: `import.inc.php` also picked up some unrelated SQL indentation tweaks. Not a blocker — just worth flagging since they're outside the PHP 8 scope.

Everything else looks clean. Once the typo is fixed this is ready.

Wachhund:
Quick correction on my previous note about `import.inc.php` — I took a closer look at the SQL indentation changes and they're actually a cleanup, not noise.

Both query blocks in that file were inconsistently indented before:

- The first one had the SQL continuation lines (`where`/`and`/`order`) one level deeper than the opening `"select`
- The second one had them shallower than the opening `"select`

Your version normalizes both to "continuation lines at the same indent as the opening string", which matches the LiteCart SQL convention shown in CLAUDE.md. So that part is fine — disregard my scope comment on this file.

Only the `srt_contains` typo in `ent_image.inc.php` still needs fixing.

s22-tech:
Thanks for catching the typo!

timint:
I like this. It's the right way to go when drawing the line at PHP 8.0. Thank you 👍

PR #518, opened 2026-05-14
Switching to named parameters

s22-tech:
If this is acceptable, I'll start working on other pages as well.

### DIFF
--- a/public_html/frontend/templates/default/pages/contact.inc.php
+++ b/public_html/frontend/templates/default/pages/contact.inc.php
@@ -9,42 +9,70 @@
 
 					<h1><?php echo t('title_contact_us', 'Contact Us'); ?></h1>
 
-					<?php echo f::form_begin('contact_form', 'post', null, true); ?>
+					<?php echo f::form_begin(
+						name      : 'contact_form',
+						method    : 'post',
+						action    : null,
+						multipart : true
+					); ?>
 
 						<div class="grid">
 							<div class="col-md-6">
 								<label class="form-group">
 									<div class="form-label"><?php echo t('title_firstname', 'First Name'); ?></div>
-									<?php echo f::form_input_text('firstname', true, 'required'); ?>
+									<?php echo f::form_input_text(
+										name       : 'firstname',
+										input      : true,
+										parameters : 'required',
+									); ?>
 								</label>
 							</div>
 
 							<div class="col-md-6">
 								<label class="form-group">
 									<div class="form-label"><?php echo t('title_lastname', 'Last Name'); ?></div>
-									<?php echo f::form_input_text('lastname', true, 'required'); ?>
+									<?php echo f::form_input_text(
+										name       : 'lastname',
+										input      : true,
+										parameters : 'required',
+									); ?>
 								</label>
 							</div>
 						</div>
 
 						<label class="form-group">
 							<div class="form-label"><?php echo t('title_email_address', 'Email Address'); ?></div>
-							<?php echo f::form_input_email('email', true, 'required'); ?>
+							<?php echo f::form_input_email(
+								name       : 'email',
+								input      : true,
+								parameters : 'required',
+							); ?>
 						</label>
 
 						<label class="form-group">
 							<div class="form-label"><?php echo t('title_subject', 'Subject'); ?></div>
-							<?php echo f::form_input_text('subject', true, 'required'); ?>
+							<?php echo f::form_input_text(
+								name       : 'subject',
+								input      : true,
+								parameters : 'required',
+							); ?>
 						</label>
 
 						<label class="form-group">
 							<div class="form-label"><?php echo t('title_message', 'Message'); ?></div>
-							<?php echo f::form_textarea('message', true, 'required style="height: 250px;"'); ?>
+							<?php echo f::form_textarea(
+								name       : 'message',
+								input      : true,
+								parameters : 'required style="height: 250px;"',
+							); ?>
 						</label>
 
 						<label class="form-group">
 							<div class="form-label"><?php echo t('title_attachments', 'Attachments'); ?></div>
-							<?php echo f::form_input_file('attachments[]', 'multiple accept=".jpg,.jpeg,.png,.gif,.webp,.avif,.txt,.doc,.docx,.pdf,.mp4"'); ?>
+							<?php echo f::form_input_file(
+								name       : 'attachments[]',
+								parameters : 'multiple accept=".jpg,.jpeg,.png,.gif,.webp,.avif,.txt,.doc,.docx,.pdf,.mp4"',
+							); ?>
 						</label>
 
 						<?php if (settings::get('captcha_enabled')) { ?>
@@ -55,7 +83,12 @@
 						<?php } ?>
 
 						<div>
-							<?php echo f::form_button('send', t('title_send', 'Send'), 'submit', 'style="font-weight: bold;"'); ?>
+							<?php echo f::form_button(
+								name       : 'send',
+								value      : t('title_send', 'Send'),
+								type       : 'submit',
+								parameters : 'style="font-weight: bold;"',
+							); ?>
 						</div>
 
 					<?php echo f::form_end(); ?>


Wachhund:
Thanks for asking before going wide on this — appreciated, because the trade-offs here aren't obvious.

Named parameters definitely have upsides. A nice side effect: they'd let the deprecated argument-swap hacks in `func_form.inc.php` (the `form_input_money` / `form_regional` / `form_toggle` etc. blocks that do `list($a, $b) = [$b, $a]` based on regex matches) go away cleanly.

That said, a few concerns worth weighing before this scales out:

**Template scale.** A quick grep shows `f::form_input_text(` alone is called in ~47 files at ~257 sites across v3 templates. Each single-line call becomes a 5-line block, so we're looking at thousands of added lines for templates whose main virtue is being scannable. This is the main worry.

**Template stability for addons.** v3 is still pre-release so there's no broad addon ecosystem patching templates via vMod yet — so this isn't a "breaks everything today" concern. But once 3.0 ships, addon authors will rely on stable template code for their `<find>` blocks. So if a stylistic switch like this is going to happen, doing it *before* 3.0 is the right window — and ideally just once, with a deliberate freeze afterward.

**Consistency.** Halfway is worse than either end — once one page goes named-args, the rest looks inconsistent until everything's converted. And there's nothing in `STANDARD.md` yet that mandates either style, so this is a project-wide convention call.

**Style nit.** The aligned colons (`name       :`) are pretty, but LiteCart's general style is compact PECL without column padding — that'd probably want to go.

My suggestion: probably worth waiting for Tim to weigh in on the strategic direction (named params project-wide vs. positional) before converting more pages — and on whether the verbosity trade-off is worth it for templates specifically.

Either way, glad you flagged this on one file first rather than dropping 50 of them. 👍

Wachhund:
One more thing worth noting on the multi-line formatting itself, separate from the named-params question:

The existing v3 codebase keeps form helpers on a single line even when the arguments get fairly involved. From Tim's own code in `backend/apps/customers/edit_customer.inc.php:243`:

```php
<?php echo f::form_button('sign_in', ['true', t('text_sign_in_as_customer', 'Sign in as customer')], 'submit', 'class="btn btn-default btn-block"'); ?>
```

That's 4 positional args, a nested array, and a `t()` call — all on one line. The recent direction in `dev-major` has also been compacting, not expanding (the "Whitespace polishing", "Replace legacy styled block comments with minimal version", `switch() → match()` commits).

So even *if* named params win out as the convention, the inline form would probably fit the project's style better:

```php
f::form_input_text(name: 'firstname', input: true, parameters: 'required');
```

A bit longer than the positional original, but still one line and scannable. The 5-line-per-call format is the part most at odds with how the rest of v3 looks.

s22-tech:
deleted

timint:
There are times when named arguments can be preferred. Pretty much when they are plenty. But in many of those cases that has been resolved by passing an array of args where some are optional some are not.

In general I agree with Wachhund on this. The nested order of args feel like what needs to be default. But when that is not optimal we should have this in mind.

In this topic let's also see some best use scenarios and see what we learn from that. 🙂

s22-tech:
Another nice thing about named arguments is that you can see the name of each parameter while in the template file without having to switch back and forth to the functions file. 

Plus, param order no longer breaks the code - you can have the params in any order you like.

Wachhund:
Picking up on Tim's call for best-use scenarios — a few thoughts on where named args earn their keep vs. where positional stays better:

### Where named args genuinely help

**1. Bare boolean flags.** This is the strongest case in our codebase:

```php
f::form_input_text('firstname', true, 'required');     // what does true mean?
f::form_input_text(name: 'firstname', input: true, parameters: 'required');
```

Hard to argue against clarity here when reading templates cold.

**2. Functions that already carry deprecation-swap hacks.** `func_form.inc.php` currently has these in `form_input_money`, `form_regional`, `form_regional_textarea`, `form_regional_wysiwyg`, `form_toggle`, `form_function`, `form_select_attribute_value`, and `form_select_zone`:

```php
list($name, $currency_code) = [$currency_code, $name];
trigger_error('Passing currency code as 1st parameter ... is deprecated...');
```

If those signatures move to named args natively, the regex-sniffing + swap dance can come out, signatures become natural, and the deprecation warnings can age out cleanly. That's real structural cleanup, not just style.

### Counter to "see param names without switching files"

That's a strong argument on its face, but modern IDEs already solve it via parameter inlay hints — PhpStorm has "Parameter name hints" on by default, and VSCode + Intelephense exposes the same. They render the call as:

```
f::form_input_text(/*name:*/ 'firstname', /*input:*/ true, /*parameters:*/ 'required')
```

So readability in templates is achievable without inflating the source. Reorder-safety remains a real benefit of named args, but in practice nobody reorders args inside a template — that's mostly a refactor-time concern.

### Hybrid pattern Tim already uses

The array-of-options approach Tim mentioned (`['input' => true, 'parameters' => 'required']`) is the de-facto pattern in WooCommerce / Magento / Symfony Forms. Worth considering for any future helper that grows past ~4 args — keeps the call site compact while making the verbose options live in an array.

### Concrete proposal for moving forward

Rather than going wide on templates, what about a focused scope:

1. Migrate **only** the deprecation-burdened `form_*` functions to native named args in their signatures
2. In the same pass, retire the swap hacks and the corresponding `trigger_error('...is deprecated...')` lines
3. Templates keep calling positional — fully compatible
4. Named args become a documented option for "many args / boolean flags" cases, not the default

Smaller diff, real cleanup win, no template bloat, and gives us a concrete reference point to learn from before deciding wider strategy. Could be a follow-up PR from this one if @s22-tech is up for it.

Either way — appreciate the discussion. The deprecation cleanup angle wouldn't have come up without this PR.

s22-tech:
@Wachhund All excellent points.  You were much more eloquent than I was.

I haven't seen the array-of-options code yet.  I searched on your sample but it didn't come up.  Can someone point me to a usage of that in LC3?  I'd like to see it in action.

This is just a personal opinion, but I think PhpStorm's parameter name hints you showed is quite ugly - handy, but ugly.  I've never seen that before, since I use BBEdit.  It's not an IDE, but it has the best file comparison window I've ever used and I can't imagine using anything else.

One followup to your point that "nobody reorders args inside a template".  That's true.  I guess the only benefit is that it would make templates easier to read.  At least people can now use named parameters in their vmods going forward, so that's a good thing.

And one last point about named params making the code longer.  Sometimes it's better to pay a little extra for full service at the gas station than it is to pump it yourself.  `=:D`

O.K. That's my 2¢.  I'll leave it to the powers that be.

timint:
Just to see all the variations one can use for arguments.

```php
add_item($id, $name, $qty, ...);

add_item([
  'quantity' => $qty,
  'name' => $name,
  'id' => $id,
  ...
]);

add_item(quantity: $qty, name: $name, id: $id, ...);

add_item(
  quantity: $qty,
  name: $name,
  id: $id,
  ...
);
```

The first  two variations is what can be found today. The last two from this proposal.

`quantity: $qty, name: $name, id: $id` is naming the names to me. We don't need that. But I understand they would be helpful when not passing a named variable.

What's related to this but is also a solution; is to stop using the f:: wrapper and instead go with `use` so IDEs knowingly can reference the helper functions. I haven't thought of that before as I know the platform from the back of my head, but being able to reference them would be very helpful to developers. So if there was a simple way to teach the IDE where to reference the f:: helper functions I would wanna look into that. The whole idea with f:: was to not have to deal with includes|use|require.

s22-tech:
O.K.  I can see how the array solution would be better than non-named parameters, but to be honest, that actually involves MORE typing than named params.  I don't see the advantage.

Your 4th variation is the cleanest looking method of all of them, even if some functions "name the names".  There's no guessing doing it that way.

And please don't get rid of `f::` for `use` - that would seem like a giant step backwards.  I love that it automatically loads the proper func file when needed.  Elegant simplicity!

timint:
We also have `add_item($item)` which is the already assembled array as minimalistic as it can be.

🙂 I like f:: a lot too. I won't give it up without a fight. But I am aware of it's disadvantage.

To revert back to the use case in this PR. That example is a no go for me in terms of anti-minimalism. The extra whitespace bytes and keywords are descriptive. But the downsides doesn't weight it up enough to adapt.

Wachhund:
Understood, Tim — the anti-minimalism call makes sense for templates, and the 4-variation rundown nicely sums up where we landed.

Bigger picture though: this discussion shook loose two follow-up threads that wouldn't have surfaced otherwise — #519 (the `$parameters` array idea) and #520 (the type-objects you've had on the back burner). Both feel more strategically interesting than the named-args-everywhere version of this PR.

@s22-tech — even if this one doesn't ship, you kicked off the right conversation. The deprecation-hacks angle and the array-of-options pattern came directly out of poking at this. Hope you'll stick around for #519 — that's where the real cleanup payoff is.

Happy to see this PR closed once we move forward on those.

timint:
Closing, but keep the ideas coming. :)

PR #521, opened 2026-05-15
[v3] Output escape audit — close XSS-001/002/004/006 + search.js DOM hardening

Wachhund:
Picking up the four XSS paths from #480 that didn't quite make it into the spring hardening sweep — plus a defense-in-depth pass on the backend search.js DOM builder while I was in the file anyway.

## Summary

| Path | Severity | What's hardened |
|------|----------|------------------|
| **XSS-001** Formatted addresses | HIGH | 7 sites now use `nl2br(f::escape_html(f::format_address(...)))`. Escape runs **before** nl2br so the `<br />` tags survive. |
| **XSS-002** Admin customer list | HIGH (**unauth** via reverse-DNS PTR) | `company`, `email`, `last_hostname`, `group_name`, `firstname+lastname` all wrapped. Stray `</a>` markup in the email column fixed too. |
| **XSS-003** Backend search DOM | HIGH | `search.js` result builder rewritten — `.text()` for content, `.attr()` for attributes, `.css()` for inline style, and the `'ul[data-group="...'+ group.name +'"]'` selector argument is now a stored `$ul` reference. |
| **XSS-004** Email lists | HIGH (**unauth** via contact form) | `subject` and recipient-name columns escaped in sent / failed / scheduled / drafts. drafts `(Untitled)` fallback markup preserved. `ent_email::set_subject()` additionally strips `<script>`/`<style>` case-insensitively as defense in depth. |
| **XSS-006** Order contact / tax | HIGH | `email`, `phone`, `tax_id` escaped at backend order detail, frontend checkout review, and printable packing slip. |
| **+ defense-in-depth** search-result URL | (info) | `safe_link()` allowlist on `result.link` — `http(s):`, `mailto:`, root-relative `/`, scheme-less paths pass through. `javascript:`, `data:`, `vbscript:`, protocol-relative `//evil` → `#`. Today all providers ship `document::ilink()`, this guards future providers. |

## Notes

- `func_format.inc.php` `format_address()` docblock now documents that the return is not HTML-safe and that the canonical render pattern requires `f::escape_html()` before `nl2br()`. Ordering matters — the reverse escapes the `<br />` tags too.
- `last_hostname` is filled from `gethostbyaddr()` at three call sites (admin, customer, route). Render-layer escape catches it; PTR-injection no longer ships HTML into admin views.
- Email-subject defense-in-depth strip runs **before** the existing CRLF/tab sanitiser so it slots in cleanly.

## Test plan

[...]

Issue #516, opened 2026-05-11
Installer broken on dev-major after the install/ restructure


Wachhund:
Was going to take a swing at the CLI installer this week — turns out the install/ tree picked up some changes while I wasn't looking. :)

Here's what I ran into:

## Gap inventory

| File / line | What |
|---|---|
| `install.php:14`, `upgrade.php:18` | `require_once 'includes/bootstrap.inc.php'` — file removed in 6ba4361 |
| `install.php:15`, `upgrade.php:19` | `require_once 'includes/security_headers.inc.php'` — same |
| `install.php:18`, `upgrade.php:109` | `install_is_cli()` — was defined in the deleted `bootstrap.inc.php`, no replacement landed |
| `install.php:27`, `upgrade.php:126` | `install_send_security_headers()` — replaced by `csp_send_headers()` in `functions.inc.php`, but the call sites still use the old name |
| `bootstrap.inc.php:9-21` (deleted) | CLI polyfill (`$_SERVER['SERVER_SOFTWARE']='CLI'` + `getopt() → $_REQUEST`) — didn't get carried over into `init.inc.php` |
| `init.inc.php:46-49` | new unconditional `install_is_locked()` check at include-time — fine for install.php (which already opts in at l.26-29), but blocks upgrade.php, which has to run when the lock is present |

Net effect: both web and CLI fatal on the missing require before any output. CI stays green because `platform-tests` runs against a pre-populated DB and skips the install/upgrade entry points.

## Repro

```bash
git checkout dev-major
docker-compose up -d --build
# http://localhost:9080/install/ → fill form → "Install Now"
# Fatal error: require_once(): Failed opening '.../install/includes/bootstrap.inc.php'
```

CLI: `php install/install.php --help` fatals on the same line.

## How I'd adjust it

1. Pull the CLI polyfill (the `$_SERVER` block + `getopt('', $INSTALL_CLI_OPTIONS)` into `$_REQUEST`) back into `init.inc.php`.
2. Move `install_is_cli()` into `init.inc.php` too (was at `bootstrap.inc.php:42-50`).
3. Drop the unconditional lock check at `init.inc.php:46-49` — `install.php:26-29` already does it explicitly, and `upgrade.php` needs to be allowed when the lock is present.
4. Either alias `install_send_security_headers() → csp_send_headers()` in `functions.inc.php`, or rename the two call sites. Alias has less surface.
[...]

PR #522, opened 2026-05-16
[v3] Restore broken installer/upgrader requires (closes #523)

### LINKED ISSUE #523: [v3] Restore broken installer/upgrader requires (function-level fix)
Function-level slice of #516 — restore the four broken behaviours so install / upgrade entry points work again. No file-layout changes; that's #524.

## Scope

Re-land what `6ba4361` removed when it dropped `bootstrap.inc.php` and `security_headers.inc.php`, folded into the files that survived rather than re-creating the deleted ones:

| Gap | Where it lives now |
|-----|--------------------|
| `install.php:14` / `upgrade.php:18` requiring deleted `bootstrap.inc.php` | swapped to `init.inc.php` |
| `install.php:15` / `upgrade.php:19` requiring deleted `security_headers.inc.php` | folded into `functions.inc.php` (alias on `csp_send_headers()`) |
| `install_is_cli()` missing — used by `install.php:18`, `upgrade.php:109` | new function in `functions.inc.php` |
| `install_send_security_headers()` missing — used by `install.php:27`, `upgrade.php:126` | backward-compat alias in `functions.inc.php` |
| CLI polyfill (`$_SERVER` normalize + `getopt` into `$_REQUEST`) | re-added in `init.inc.php` at the top |
| Unconditional `install_is_locked()` gate at `init.inc.php:46-49` blocking `upgrade.php` | removed from the shared bootstrap — `install.php` already gates it explicitly, and `upgrade.php` must run when the lock is present |

## Out of scope (covered by sibling issue)

The two-tier split into `includes/NN-*.inc.php` + `pipelines/{install,upgrade}/NN-*.inc.php` is a structural change that benefits from landing on a working baseline. Tracked separately so this one stays surgical and revertable.

## Verification

- `php public_html/install/upgrade.php --help` renders the banner (was fataling on line 18)
- `php public_html/install/install.php --document_root=… --db_server=…` walks through parameter / PHP-version / extension / display_errors checks before the storage step
- `php -l` clean on all four touched files (PHP 8.5)
- CI on a clean baseline runs through install + platform + e2e on both DB matrix columns

## PR

#522 — diff is ~80 lines, four files, no behaviour change for the web installer.

Wachhund:
Picking up #523 (sub-issue of #516) — `6ba4361` (remove remaining files from moved code) deleted `install/includes/bootstrap.inc.php` and `security_headers.inc.php`, but `install.php:14-15` and `upgrade.php:18-19` still required them. Both entry points fatal before any output. Pre-existing on every CI run since the cleanup landed.

## What's folded back where

| Removed helper | Now lives in | Notes |
|---------------|--------------|-------|
| CLI polyfill (`$_SERVER` normalize + `getopt()` → `$_REQUEST`) | `init.inc.php` (early) | Reads `$INSTALL_CLI_OPTIONS` if the caller set it, falls back to empty. Web requests untouched. |
| `install_is_cli()` | `functions.inc.php` | Single SAPI check against the polyfilled `$_SERVER['SERVER_SOFTWARE']`. |
| `install_send_security_headers()` | `functions.inc.php` | Thin alias around `csp_send_headers()` so the two existing call sites keep working without renames. |
| `FS_DIR_APP` / `FS_DIR_STORAGE` | `init.inc.php` (was already there) | unchanged. |

## Side adjustments

| Change | Why |
|--------|-----|
| `init.inc.php` no longer hardcodes `BACKEND_ALIAS` / `DB_TABLE_PREFIX` | install.php sets them from user-supplied parameters and upgrade.php picks them up from `storage/config.inc.php`. Keeping the early defaults caused `Constant already defined` warnings on every install run. |
| `init.inc.php` no longer enforces the lock at include time | install.php gates it explicitly at line 26-29 already, and upgrade.php must run *while* the lock is present — that's its job. A shared init can't make that call. |
| `init.inc.php` defines `DOCUMENT_ROOT` when either `$_SERVER` provides it or `--document_root` was passed | CLI runs no longer crash on the missing server variable. |
| `install.php` guards its own `DOCUMENT_ROOT` / `VMOD_DISABLED` defines with `!defined()` | Avoids "already defined" warnings now that init.inc.php may have set them first. |
| `upgrade.php` drops the standalone `compatibility.inc.php` require | init.inc.php already pulls it. |

## Verified locally

- `php public_html/install/upgrade.php --help` renders the banner cleanly (was fataling on line 18 before).
- `php public_html/install/install.php --db_server=... --document_root=public_html ...` walks through parameter / PHP-version / extension / display_errors checks before failing on the missing local DB (was fataling on line 14 before).
- `php -l` clean on all four touched files (PHP 8.5).

Two pre-existing CLI warnings remain (undefined `$_SERVER['SERVER_PROTOCOL']` / `SERVER_NAME` in `compatibility.inc.php:48,52`) — out of scope here, but worth a note if you want them addressed in a follow-up.

## Test plan

- [ ] CI green — install / platform / e2e on both DB matrix columns
- [ ] Web installer reachable at `/install/` after a Docker rebuild
- [ ] `php install/install.php --help` shows usage from CLI

Once this lands, #521 ([v3] Output escape audit) can re-run CI on a clean baseline. Happy to rebase that one on top if it helps.

Closes #523.

Parent #516 stays open while #524 (the NN-numbered submodule split) is sequenced on top of this baseline.

### DIFF
--- a/public_html/install/includes/functions.inc.php
+++ b/public_html/install/includes/functions.inc.php
@@ -2,6 +2,26 @@
 
 	include_once __DIR__.'/../../includes/functions/func_file.inc.php';
 
+	/*
+		Returns true when the current process is a CLI invocation. The CLI
+		polyfill in init.inc.php normalises $_SERVER['SERVER_SOFTWARE'] to
+		'CLI' before this function is reachable, so a single comparison
+		suffices. Defense-in-depth against mis-configured FPM pools is left
+		to the SAPI check; entry points may also gate on php_sapi_name().
+	*/
+	function install_is_cli() {
+		return ($_SERVER['SERVER_SOFTWARE'] ?? '') === 'CLI';
+	}
+
+	/*
+		Backwards-compatible alias for the previous standalone helper. Both
+		install.php and upgrade.php still call this name; future call sites
+		may use csp_send_headers() directly.
+	*/
+	function install_send_security_headers() {
+		csp_send_headers();
+	}
+
 	function csp_send_headers() {
 
 		header('Content-Security-Policy: ' . implode('; ', [
--- a/public_html/install/includes/init.inc.php
+++ b/public_html/install/includes/init.inc.php
@@ -5,12 +5,31 @@
 	mb_internal_encoding('UTF-8');
 	mb_http_output('UTF-8');
 
-	define('BACKEND_ALIAS', 'admin');
-	define('DB_TABLE_PREFIX', 'lc_');
+	// BACKEND_ALIAS / DB_TABLE_PREFIX are NOT defined here — install.php
+	// defines them from user-supplied parameters (line ~181), and upgrade.php
+	// picks them up from the existing storage/config.inc.php.
+
+	// CLI polyfill: normalise $_SERVER and collect getopt into $_REQUEST.
+	// Entry points may pass their own long-options list via $INSTALL_CLI_OPTIONS
+	// before including this file; fall back to an empty list.
+	if (!isset($_SERVER['REQUEST_METHOD'])) {
+		$_SERVER['SERVER_SOFTWARE'] = 'CLI';
+		$_SERVER['REQUEST_METHOD'] = 'GET';
+		$_SERVER['REMOTE_ADDR'] = '127.0.0.1';
+
+		if (!empty($INSTALL_CLI_OPTIONS) && is_array($INSTALL_CLI_OPTIONS)) {
+			$_REQUEST = getopt('', $INSTALL_CLI_OPTIONS);
+		}
+	}
 
 	// Filesystem constants (redefined safely if already set).
 	if (!defined('DOCUMENT_ROOT')) {
-		define('DOCUMENT_ROOT', rtrim(str_replace('\\', '/', realpath($_SERVER['DOCUMENT_ROOT'])), '/'));
+		if (!empty($_SERVER['DOCUMENT_ROOT'])) {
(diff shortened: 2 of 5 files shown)


Wachhund:
Quick CI status — installer side is good, surfacing one unrelated thing on the matrix.

| Job | Result | Notes |
|-----|--------|-------|
| PHP & Asset Linting | ✅ | SCSS-Lint script fix lands here too (the `RUNNER[@]` / `$file` bash bug from d6fc6281d) — happy to split if you'd rather keep this PR scoped tight. |
| Gulp Build | ✅ | |
| Install Application (mysql:latest) | ✅ | Installer runs through end-to-end. This is the verification the function-level fix was after. |
| Install Application (mariadb:latest) | ❌ | Container reports `unhealthy` after the 10×10s health-retries window. Not caused by this PR — looks like a matrix env-var mismatch: `mariadb:latest` may have dropped the `MYSQL_*` legacy aliases that PROJ-31 (#512) relied on. Worth a quick look at the service block in `ci.yml`. |
| Platform Tests | ⏸️ | Gated on Install — MySQL column would run if Install split were independent, but `needs:` is at job level. |
| E2E Tests (Playwright) | ⏸️ | Same. |

Net: the install/upgrade requires regression from #523 is closed by this PR. The MariaDB matrix wobble is independent — happy to send a small follow-up that adds `MARIADB_*` env vars alongside `MYSQL_*` if you want me to take that one too, or leave it for whatever cadence works.

Also nudging visibility on #524 (the structural split into NN-numbered submodules) — same parent issue, sequenced after this one so it lands on a working baseline.

timint:
Thank you. Just wait till I break it again 😄😄

PR #525, opened 2026-05-16
[v3] Fix CI MariaDB matrix — MARIADB_* env vars and mariadb-admin healthcheck

### LINKED ISSUE #523: [v3] Restore broken installer/upgrader requires (function-level fix)

Wachhund:
Tiny CI-only fix — the matrix moved from `mariadb:12` (PROJ-31, #512) to `mariadb:latest` somewhere along the way, and current MariaDB releases stopped accepting the legacy `MYSQL_*` env-var aliases that the service block was relying on. Result: every run on `dev-major` has the MariaDB column failing at `Initialize containers` with `unhealthy` after the retries window. The MySQL column is fine.

## What changes

| Block | Before | After |
|-------|--------|-------|
| Install / platform-tests / e2e-tests env vars | `MYSQL_*` only | `MYSQL_*` **and** `MARIADB_*` set side by side |
| Install / e2e-tests matrix healthcheck (MariaDB row) | `mysqladmin ping -h localhost` | `mariadb-admin ping -h localhost` |
| Install / e2e-tests matrix healthcheck (MySQL row) | unchanged | unchanged (`mysqladmin`) |
| platform-tests hardcoded MariaDB service healthcheck | `mysqladmin ping` | `mariadb-admin ping` |

`mariadb-admin` has shipped in every MariaDB release since 10.5 — the `mysqladmin` symlink is being phased out and is already unreliable on `:latest`. MySQL's tool keeps the original name.

Both env-var families are set rather than picking one, so the block survives whichever engine drops its alias next. Cost is zero — each engine ignores the other family.

## Verified

- `python -c 'import yaml; yaml.safe_load(open(".github/workflows/ci.yml"))'` — green
- ci.yml diff is +22/-3, single file
- No application-code or schema change

## Test plan

- [ ] CI green across all four matrix columns (`install` mariadb-latest + mysql-latest, `e2e-tests` mariadb-latest + mysql-latest)
- [ ] `platform-tests` job runs through against the hardcoded MariaDB service

Sequenced behind #522 (`closes #523`) only because that one is the substantive change — this one is independent and merge-order doesn't matter.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,7 +110,7 @@ jobs:
         db:
           - image: mariadb:latest
             slug: mariadb-latest
-            health: 'mysqladmin ping -h localhost'
+            health: 'mariadb-admin ping -h localhost'
           - image: mysql:latest
             slug: mysql-latest
             health: 'mysqladmin ping -h localhost'
@@ -119,10 +119,17 @@ jobs:
       database:
         image: ${{ matrix.db.image }}
         env:
+          # Both env-var families are set so the service block works on
+          # mariadb:latest (which no longer accepts the legacy MYSQL_*
+          # aliases) and on mysql:latest (which ignores MARIADB_*).
           MYSQL_ROOT_PASSWORD: litecart
           MYSQL_DATABASE: litecart
           MYSQL_USER: litecart
           MYSQL_PASSWORD: litecart
+          MARIADB_ROOT_PASSWORD: litecart
+          MARIADB_DATABASE: litecart
+          MARIADB_USER: litecart
+          MARIADB_PASSWORD: litecart
         ports:
           - 3306:3306
         options: >-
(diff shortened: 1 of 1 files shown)


Wachhund:
Status note — CI failed on the SCSS Lint step (`scripts/scan-scss-syntax.sh` bash bug from d6fc6281d), not on anything this PR touches. Same preexisting block that's hitting every PR on `dev-major`. #522 carries the SCSS script fix as a side commit; once that lands this PR will rebase clean.

Merge sequence that drains everything: #522 first (closes #523 + scrubs the SCSS script), then this one runs through, then #521 catches the XSS work on a green baseline. No coupling between #525 and #522 in the diffs themselves — just CI ordering.

Issue #529, opened 2026-05-16
[v3] Add type_translation datatype


Wachhund:
Sub-issue of #520 — `type_translation` for multilingual payloads, sequenced after `type_money` (#528) and `type_email` (#526).

## Why it earns its weight

Regional fields in v3 already store per-language values as JSON arrays. Today the calling code has to manually pick the right locale on render:

```php
// Today
$row = database::query("...")->fetch();
$name_payload = json_decode($row['name'], true);
echo $name_payload[language::$selected['code']] ?? $name_payload['en'] ?? '';
```

That `?? $name_payload['en'] ?? ''` fallback dance gets re-implemented at every render site (admin product pages, frontend category listings, printable copies, JSON endpoints). A `type_translation` collapses it.

```php
$name = new type_translation([
  'en' => 'Rubber Ducks',
  'de' => 'Gummienten',
  'sv' => 'Gummiankor',
]);

echo $name;                      // current locale (language::$selected['code'])
$name->in('de');                 // 'Gummienten'
$name->jsonSerialize();          // ['en' => 'Rubber Ducks', 'de' => 'Gummienten', 'sv' => 'Gummiankor']
$name->set('fr', 'Canards');     // fluent setter

isset($name['de']);              // true (ArrayAccess for back-compat)
```

## Shape proposal (analog to type_url / type_email / type_money)

- `__construct(array|string|type_translation $input, ?string $default_locale = null)`
  - `array` of locale → string → full payload
  - `string` → seed the current locale with that value
  - `type_translation` → copy
- Magic getter: `value` (current locale, with fallback chain)
- Methods:
  - `in(string $locale): string` — explicit locale lookup with fallback
  - `set(string $locale, string $value): self` — fluent
  - `has(string $locale): bool`
  - `locales(): array` — all populated locales
- `__toString()` — renders the active locale via `language::$selected['code']`, falls back to store default then to any populated locale
- `jsonSerialize()` — full `[locale => value]` payload, matches the existing JSON column shape
- `ArrayAccess` — `$translation['de']` is shorthand for `$translation->in('de')`, makes existing call sites that do `$row['name']['de']` keep working

## What it lets us delete or simplify

| Where | Pattern that goes away |
|-------|------------------------|
| Frontend product / category / brand templates | `$row['name'][language::$selected['code']] ?? $row['name']['en'] ?? ''` |
| `form_regional_*` helpers and their fallback chains | Render side gets a value object instead of an array |
[...]

PR #530, opened 2026-05-16
[v3] Add type_money datatype with dual construction modes (#528)

### LINKED ISSUE #528: [v3] Add type_money datatype
Sub-issue of #520 — `type_money` as the next datatype to land after `type_email` (#526).

## Why it earns its weight

Currency math + locale-aware formatting + conversion are exactly what value objects were invented for. The boilerplate is scattered: today the codebase does `currency::format()` / `currency::convert()` / `number_format(...)` in dozens of places, each repeating the rounding/currency-pair tracking by hand. A `type_money` value object collapses that.

```php
$price = new type_money(99.99, 'USD');
$price->amount;            // 99.99
$price->currency;          // 'USD'

$converted = $price->convert('EUR');           // → new type_money
$gross     = $price->gross($tax_rate);         // → new type_money
$net       = $price->net($tax_rate);           // → new type_money

echo $price;               // '$99.99' (currency::format under the hood)
$price->jsonSerialize();   // ['amount' => 99.99, 'currency' => 'USD']
```

Immutable-style methods (each transformation returns a fresh `type_money`) keep call sites predictable and ergonomic for chaining: `$price->convert('EUR')->gross($rate)`.

## What it lets us delete or simplify

A non-exhaustive list of patterns that today carry the rounding/currency state by hand:

| Where | Pattern |
|-------|---------|
| `backend/apps/orders/order.inc.php` | `currency::format($order['total'], false, $order['currency_code'], $order['currency_value'])` × ~6 sites |
| `frontend/templates/default/pages/printable_order_copy.inc.php` | Same pattern for item.price, item.tax, item.sum × N |
| `frontend/templates/default/pages/checkout/index.inc.php` | Same |
| `backend/apps/catalog/edit_product.inc.php` | Price field rendering with explicit currency_code |
| Multiple order_total modules | Internal arithmetic carries (`amount`, `currency_code`, `currency_value`) tuples |

The follow-up entity refactor (separate PR) can swap raw decimals → `type_money` instances in `ent_order::data['items'][N]['price']` etc., once the value type lands.

## Shape proposal (analog to type_url / type_email)

- `__construct(float|string|array|type_money $input, ?string $currency = null)`
  - `float`/`string` + `$currency` → `(amount, currency)`
  - `array` with `['amount' => …, 'currency' => …]` → drop-in for any existing JSON shape
  - another `type_money` → copy
- Magic getters: `amount`, `currency`, `formatted`, `is_zero`
- Setters: `amount`, `currency` (each returns `$this` for fluent style, but doesn't mutate prior references — see "Mutability" below)
- Math (returns new `type_money`):
  - `convert(string $target_currency)` — uses `currency::convert()`
  - `add(type_money $other)` — same-currency only, throws otherwise
  - `subtract(type_money $other)` — same
  - `multiply(float $factor)` — for quantity
  - `gross(float $tax_rate)` / `net(float $tax_rate)`
- `__toString()` — `currency::format($amount, true, $currency)`
- `jsonSerialize()` — `['amount' => $amount, 'currency' => $currency]`

### Mutability question

Two options worth nailing down before implementation:

- **A) Mutable**: setters change the object, math methods return new objects. Matches `type_url` (where `$url->host = 'x'` mutates).
- **B) Fully immutable**: setters return a new object. Forces `$price = $price->with_currency('EUR')` instead of `$price->currency = 'EUR'`.

`type_url` chose mutable. Following the existing pattern is the cheaper call unless there's a concrete reason to diverge — probably go A.

## Precision

Use string-based math via `bcmath` for `add`/`subtract`/`multiply` to avoid float drift, then cast back to float for `convert` (which always involves a non-exact rate anyway). Internal storage: float, because that's how `currency::format` consumes it. Trade-off documented in the class.

## Scope

| In scope | Out of scope (separate PR) |
|----------|----------------------------|
| `public_html/includes/datatypes/type_money.inc.php` | Migrating call sites to use `type_money` |
| `tests/type_money.php` (precision, conversion, format, tax math) | Refactoring `ent_order` to store `type_money` instances |
| Reuse of existing `currency::format()` / `currency::convert()` | Removing `currency::format` callers wholesale |

## Verification plan

- Construction from all four input shapes round-trips correctly
- `convert()` matches `currency::convert()` output for known pairs
- `add()` / `subtract()` reject cross-currency pairs
- `multiply(0)` produces a zero of the same currency
- `gross($r)->net($r)` ≈ original within `bcmath` precision
- `jsonSerialize()` shape stays stable for ent_* storage

Happy to take this as a PR after `type_email` (#527) lands.

Wachhund:
Picking up #528 — builds on the draft you posted in the types.zip dump and extends it along the lines we discussed in the comment thread.

## What changed vs. the draft

| Area | Draft | This PR |
|------|-------|---------|
| Construction | `(float, currency_code)` only | Same **plus** `(array $currency_map)` for fixed-amounts-per-currency mode |
| `convert()` return | `float` | `type_money` — the chain pattern you said you liked works now (`$price->convert('EUR')->format()`) |
| Internal storage key | `settings::get('site_currency_code')` (doesn't exist in v3) | `store_currency_code` |
| Immutability | implicit (public properties allowed mutation) | enforced — magic setter rejects with `E_USER_WARNING`, derived values come from `convert()` / `with_amount()` |
| JsonSerializable | absent | full round-trip — `mode` + `origin_currency` + `amounts` |

## The two modes

```php
// Mode A: single amount + currency, auto-converts to store currency internally.
// Use for calculated values (totals, taxes, discounts).
$total = new type_money(99.99, 'USD');
$total->in('EUR');             // float, converted on demand
$total->convert('EUR')         // new type_money in EUR
      ->format();              // 'EUR 92.05' (or whatever the rate yields)

// Mode B: fixed amounts per currency, verbatim merchant prices.
// Use for catalog prices that should NOT drift with FX-rate refreshes.
$price = new type_money([
    'USD' => 99.99,
    'EUR' => 92.00,
    'SEK' => 1050,
]);
$price->in('EUR');             // 92.00 — exact merchant price, no conversion
$price->in('GBP');             // falls back to converting from the store-currency entry
```

Detection lives in the constructor — `is_array($input)` → fixed mode, scalar → converted mode. The two modes coexist on the same type without a separate class, which keeps `type_taxable_amount` (#528 sibling) able to compose either flavour.

## Why immutable

`type_url` is mutable (you can set `$url->host = 'x'` directly), but money flows through many more code paths — totals, taxes, discounts, conversions — and a mutable instance there means spooky-action-at-a-distance when a `format()` call somewhere accidentally rewrites the price. Locking mutation to `convert()` / `with_amount()` keeps the chain explicit. Happy to discuss if you'd rather match `type_url`'s mutability, but the divergence felt worth keeping.

## Catalog-price drift, for context

The fixed mode addresses an old pain — today multi-currency catalog prices drift every time the FX cache refreshes, because each render re-runs the conversion. With `type_money` knowing it's holding a fixed merchant price, that path stops re-computing.

## Test plan

- [ ] CI green across all four matrix columns
- [ ] `tests/type_money.php` — 12 TCs:
  - TC-01 to TC-03: both construction modes + fallback conversion when a currency isn't in the fixed map
  - TC-04 to TC-05: `convert()` and `with_amount()` return fresh instances (chainable)
  - TC-06: copy constructor clones all components
  - TC-07: `jsonSerialize` round-trips through `json_encode` / `json_decode`
  - TC-08: magic setter rejects mutation with `E_USER_WARNING`
  - TC-09: `__toString` defaults to `format()` and yields a non-empty rendered string
  - TC-10: unknown component access triggers `E_USER_WARNING`
  - TC-11: `currencies` getter exposes the stored currency list
  - TC-12: zero amount is preserved and formats to a non-empty string
- [ ] Manual smoke: round-tripping a fixed-map through `ent_campaign`-style JSON storage preserves the explicit per-currency values

Closes #528.

`type_translation` (#529) comes next as a separate PR — keeping that one independent so the loop-bug fix doesn't tangle with the dual-mode discussion here.

### DIFF
--- a/public_html/includes/datatypes/type_money.inc.php
+++ b/public_html/includes/datatypes/type_money.inc.php
@@ -1,0 +1,181 @@
+<?php
+
+	/*
+		Represents a monetary amount with currency awareness. Two construction modes:
+
+		1) Single amount + currency — internally normalised to the store currency
+		   and converted on demand. Use this for calculated values (totals, taxes).
+
+		    $price = new type_money(99.99, 'USD');
+		    echo $price->in('EUR');             // float, converted from USD
+		    echo $price;                        // 'USD 99.99' formatted
+		    $price->convert('EUR')->format();   // chainable — returns new type_money
+
+		2) Fixed amounts per currency — explicit merchant-set prices that should
+		   NOT drift with FX-rate refreshes. Use this for catalog prices.
+
+		    $price = new type_money(['USD' => 99.99, 'EUR' => 92.00, 'SEK' => 1050]);
+		    echo $price->in('EUR');             // 92.00 — verbatim, no conversion
+		    echo $price->in('GBP');             // converts from store-currency entry
+
+		Drop-in for entity JSON columns that already store per-currency maps —
+		jsonSerialize() returns ['amounts' => [...], 'mode' => 'fixed'|'converted'].
+	*/
+
+	class type_money implements \JsonSerializable {
+
+		const MODE_CONVERTED = 'converted';
+		const MODE_FIXED     = 'fixed';
+
+		private $_amounts = [];
+		private $_mode;
+		private $_origin_currency;
+
+		public function __construct($input = 0, ?string $currency_code = null) {
+
+			if ($input instanceof type_money) {
+				$this->_amounts         = $input->_amounts;
+				$this->_mode            = $input->_mode;
+				$this->_origin_currency = $input->_origin_currency;
+				return;
+			}
+
+			$store_currency = (string)settings::get('store_currency_code');
+
+			if (is_array($input)) {
+
+				// Fixed-amounts-per-currency mode — explicit merchant prices, no conversion.
+				if (isset($input['amounts']) && is_array($input['amounts'])) {
+					// jsonSerialize round-trip — accept the serialised shape verbatim.
+					$this->_amounts         = array_map('floatval', $input['amounts']);
+					$this->_mode            = ($input['mode'] ?? self::MODE_FIXED) === self::MODE_CONVERTED ? self::MODE_CONVERTED : self::MODE_FIXED;
+					$this->_origin_currency = $input['origin_currency'] ?? $store_currency;
+					return;
+				}
+
+				// Plain map of currency_code => amount.
+				$this->_amounts         = array_map('floatval', $input);
+				$this->_mode            = self::MODE_FIXED;
+				$this->_origin_currency = array_key_first($this->_amounts) ?: $store_currency;
+				return;
+			}
+
+			// Single amount mode — normalise to store currency for consistent math.
+			$amount   = (float)$input;
+			$currency = $currency_code ?: $store_currency;
+			$this->_origin_currency = $currency;
+			$this->_mode            = self::MODE_CONVERTED;
+			$this->_amounts         = [
+				$store_currency => $currency === $store_currency ? $amount : (float)currency::convert($amount, $currency, $store_currency),
+			];
+		}
+
+		public function __get($name) {
+
+			switch ($name) {
+
+				case 'amount':
+					// Default to the origin currency in single-amount mode, or to the
+					// store-currency entry in fixed-map mode if present, else first.
+					return $this->in($this->_origin_currency);
+
+				case 'currency':
+				case 'currency_code':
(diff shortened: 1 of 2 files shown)


timint:
It looks like you really nailed it! Well done :)

PR #531, opened 2026-05-16
[v3] Add type_translation datatype (#529)

### LINKED ISSUE #529: [v3] Add type_translation datatype

Wachhund:
Picking up #529 — built on the draft you posted in the types.zip dump and addresses the items from the comment thread.

## What changed vs. the draft

| Area | Draft | This PR |
|------|-------|---------|
| Class / file name | `type_translations` (plural) | `type_translation` (singular, matches sibling `type_url` / `type_email` / `type_money`) |
| Fallback loop | Read `$this->_language_code` inside the loop instead of the loop var — disabled the fallback path silently | Loop now reads the local `$code`, fallback chain walks correctly |
| JsonSerializable | absent | full round-trip; returns the `{locale => string}` map verbatim, matches the JSON column shape ent_* already persists |
| ArrayAccess | absent | `$row['name']['de']`-style reads still work without forcing a sweep of every call site |
| Helpers | `set_language()` only | added `in()` / `set()` / `has()` / `locales()` |
| Empty / null safety | strict typing of locale param | always returns a string, never null — callers don't have to coalesce |

## The fallback chain (documented at `__toString`)

1. Active locale (`language::$selected['code']`)
2. Store default locale (`settings::get('store_language_code')`)
3. `'en'` as final declared fallback (LiteCart's seed locale)
4. First populated entry (defensive)
5. Empty string

Pinned in the class header so future readers don't have to reason it out from the code.

## Drop-in compatibility

`jsonSerialize()` matches the existing JSON-column shape — the entity refactor (separate PR) becomes a pass-through:

```php
// Before
$row['name'] = json_decode($row['name'], true);
echo $row['name'][language::$selected['code']] ?? $row['name']['en'] ?? '';

// After
$row['name'] = new type_translation(json_decode($row['name'], true));
echo $row['name'];   // walks the fallback chain
```

ArrayAccess means the bridging step works too:

```php
echo $row['name']['de'];   // explicit locale lookup, same fallback safety
```

## Test plan

- [ ] CI green across all four matrix columns
- [ ] `tests/type_translation.php` — 14 TCs:
  - TC-01 to TC-04: constructor variants (map / string / copy / type_translation), explicit-locale reads, active-locale `__toString`
  - TC-03: regression check for the draft's fallback-loop bug — translation containing only a non-active locale must still resolve via fallback
  - TC-05 to TC-07: string-with-locale constructor, copy clone, `set()` without disturbing existing entries
  - TC-08 to TC-09: `has()` / `locales()` filter on populated entries only
  - TC-10 to TC-11: `ArrayAccess` read and write
  - TC-12: `jsonSerialize` shape preservation + json round-trip
  - TC-13 to TC-14: empty constructor stays empty, `unset($t['de'])` removes a locale

Closes #529.

This is the second of two follow-ups to the types.zip dump. `type_money` is up in #530, sequenced ahead of this one only because the dual-mode discussion deserved the first slot. The two PRs are independent — merge in either order.

### DIFF
--- a/public_html/includes/datatypes/type_translation.inc.php
+++ b/public_html/includes/datatypes/type_translation.inc.php
@@ -1,0 +1,142 @@
+<?php
+
+	/*
+		Holds a multilingual payload (locale => string) and renders the active
+		locale on stringification with a documented fallback chain.
+
+		Example:
+		    $name = new type_translation([
+		        'en' => 'Rubber Ducks',
+		        'de' => 'Gummienten',
+		        'sv' => 'Gummiankor',
+		    ]);
+		    echo $name;            // current locale (language::$selected['code'])
+		    echo $name->in('de');  // 'Gummienten'
+		    isset($name['de']);    // true — ArrayAccess for $row['name']['de']-style callers
+
+		Fallback chain documented at __toString().
+	*/
+
+	class type_translation implements \JsonSerializable, \ArrayAccess {
+
+		private $_translations = [];
+
+		public function __construct($input = '', ?string $locale = null) {
+
+			if ($input instanceof type_translation) {
+				$this->_translations = $input->_translations;
+				return;
+			}
+
+			if (is_array($input)) {
+				// Plain {locale => string} map.
+				foreach ($input as $code => $value) {
+					if (is_string($code) && is_scalar($value)) {
+						$this->_translations[$code] = (string)$value;
+					}
+				}
+				return;
+			}
+
+			if (is_string($input) && $input !== '') {
+				// Seed the requested locale (or the active locale) with the string.
+				$code = $locale ?: ($this->_active_locale() ?: 'en');
+				$this->_translations[$code] = $input;
+			}
+		}
+
+		/*
+			Returns the value for a specific locale, walking the same fallback
+			chain as __toString. Always returns a string (never null), so callers
+			don't have to coalesce.
+		*/
+		public function in(string $locale): string {
+			if (isset($this->_translations[$locale]) && $this->_translations[$locale] !== '') {
+				return $this->_translations[$locale];
+			}
+			return $this->_fallback($locale);
+		}
+
+		public function set(string $locale, string $value): self {
+			$this->_translations[$locale] = $value;
+			return $this;
+		}
+
+		public function has(string $locale): bool {
+			return isset($this->_translations[$locale]) && $this->_translations[$locale] !== '';
+		}
+
+		public function locales(): array {
+			return array_keys(array_filter($this->_translations, fn($v) => $v !== ''));
+		}
+
+		/*
+			Fallback chain for __toString:
+			1. Active locale (language::$selected['code'])
+			2. Store default locale (settings::get('store_language_code'))
+			3. 'en' as final declared fallback (LiteCart's seed locale)
+			4. First populated entry (defensive — shouldn't happen on a sane install)
+			5. Empty string
+		*/
+		public function __toString(): string {
+			return $this->_fallback($this->_active_locale());
+		}
+
+		private function _fallback(string $requested_locale = ''): string {
+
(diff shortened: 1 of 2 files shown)


timint:
Yes!! 👍 Just perfect!

PR #538, opened 2026-05-17
[v3] Migrate form_regional_* family to accept type_translation (#537)

### LINKED ISSUE #537: [v3] Migrate form_regional_* family to accept type_translation
Sub-issue of #520 — second value-object adoption after `form_input_money` (#533/#534). Same pattern: the type-object lets us retire the regex-sniff swap-hack that lived in three of the four helpers, and brings native value-object support to all of them.

## What's in scope

| Helper | Has swap-hack today | Action |
|--------|---------------------|--------|
| `form_regional` | Yes (line 825) | Drop swap + accept `type_translation` |
| `form_regional_text` | No — already clean | Accept `type_translation` only |
| `form_regional_textarea` | Yes (line 864) | Drop swap + accept `type_translation` |
| `form_regional_wysiwyg` | Yes (line 883) | Drop swap + accept `type_translation` |

The swap-hack regex (`^[a-z]{2}$` against `$name`) is identical to what `form_input_money` had — and just like there, `rg "form_regional[a-z_]*\(['\"][a-z]{2}['\"]"` returns zero matches, so no current call site triggers the swap. Removing it costs nothing.

There's also a small pre-existing bug in `form_regional`'s deprecation message (it says "form_regional_text" instead of "form_regional") — vanishes for free when the trigger_error block is removed.

## Proposed change

Lean on the same Option-A shape that landed for `form_input_money`: repurpose `$input` to also accept a `type_translation` instance, magic-detect via `instanceof`. Smallest diff, no new signatures.

```php
function form_regional_text(\$name, \$language_code='', \$input=true, \$attributes=[]) {

  if (!\$language_code) {
    \$language_code = settings::get('store_language_code');
  }

  if (\$input instanceof type_translation) {
    \$input = \$input->in(\$language_code);
  }

  if (\$input === true) {
    \$input = form_reinsert_value(\$name);
  }
  …
}
```

`type_translation::in()` already returns a guaranteed string (no null) with the documented fallback chain, so no extra coalescing needed at the call site.

## Out of scope (leave for future iterations)

- The 44 existing call sites across 12 backend `edit_*.inc.php` files. They all pass `true` (reinsert from `$_POST`) and don't currently have a `type_translation` instance handy. Once `ent_*` entities start exposing JSON columns as `type_translation` directly (separate, larger refactor), the call sites can opt in incrementally. No need to force it in this PR.
- `form_select_zone` / `form_toggle` / `form_function` / `form_select_attribute_value` — orphans without a value-object tie-in. Unchanged.

## Verification plan

- Bare string still works (`form_regional_text('title[en]', 'en', 'Hello')`)
- `type_translation` instance accepted (`form_regional_text('title[en]', 'en', $translation)`)
- Empty `$language_code` falls back to `settings::get('store_language_code')` (existing behaviour preserved)
- No `E_USER_DEPRECATED` on the standard call patterns
- Legacy swap-hack signature no longer reorders args
- `tests/form_regional.php` modelled on `tests/form_input_money.php` covers all four helpers

Wachhund:
Second value-object adoption after [#534](https://github.com/litecart/litecart/pull/534). Same shape — drop the regex-sniff swap-hacks, accept the value-object as `\$input`, leave the call sites alone for now. `type_translation` landed in [#531](https://github.com/litecart/litecart/pull/531) so this is the natural follow-up.

Closes #537.

## What changes

| Helper | Was | Now |
|--------|-----|-----|
| `form_regional` | swap-hack + `trigger_error('…deprecated')` | `type_translation` via `instanceof` |
| `form_regional_text` | already clean | `type_translation` via `instanceof` |
| `form_regional_textarea` | swap-hack + `trigger_error('…deprecated')` | `type_translation` via `instanceof` |
| `form_regional_wysiwyg` | swap-hack + `trigger_error('…deprecated')` | `type_translation` via `instanceof` |

All four helpers now do the same thing internally:

```php
if (!\$language_code) {
  \$language_code = settings::get('store_language_code');
}

if (\$input instanceof type_translation) {
  \$input = \$input->in(\$language_code);
}
```

`type_translation::in()` already returns a guaranteed string with the documented fallback chain, so no extra coalescing at the call site.

## Risk

`rg "form_regional[a-z_]*\(['\"][a-z]{2}['\"]"` returns zero matches — no current call site passes a bare two-letter language code as `\$name`, so the swap-hacks never fired. 44 existing call sites across 12 backend `edit_*.inc.php` files all pass `true` for reinsert; they keep their behaviour verbatim. Tiny verify-only sweep.

There's also a small free pre-existing fixup tucked in: `form_regional`'s deprecation message used to say \"form_regional_text\" (wrong function name). Vanishes for free when the trigger_error block goes.

## Out of scope

- Migrating call sites to pass `type_translation` instances. None of them have one handy today — the JSON columns on `ent_*` entities are still raw arrays. When/if entities expose those as `type_translation` directly (separate, larger refactor), call sites can opt in incrementally.
- `form_select_zone` / `form_toggle` / `form_function` / `form_select_attribute_value` — same orphan-status as before, no value-object tie-in yet.

## Test plan

- [x] `php -l` clean on both touched files
- [ ] CI green (Platform Tests + Lint)
- [x] `tests/form_regional.php` covers:
  - TC-01: bare string still renders (backwards compat)
  - TC-02: `type_translation` instance accepted, cross-locale dispatch via `\$language_code`
  - TC-03: `form_regional_textarea` with `type_translation`
  - TC-04: `form_regional_wysiwyg` with `type_translation`
  - TC-05: `form_regional` (the master with `\$type` param) with `type_translation`
  - TC-06: empty `\$language_code` falls back to `store_language_code`
  - TC-07: `type_translation` fallback chain yields a non-empty value for unknown locale
  - TC-08: no `E_USER_DEPRECATED` on normal call patterns (all four helpers)
  - TC-09: legacy swap-hack signature no longer reorders args (all three helpers that had it)
  - TC-10: array attributes still merge through alongside `type_translation`

### DIFF
--- a/public_html/includes/functions/func_form.inc.php
+++ b/public_html/includes/functions/func_form.inc.php
@@ -822,13 +822,12 @@
 
 	function form_regional($name, $language_code='', $input=true, $type='text', $attributes=[]) {
 
-		if (preg_match('#^[a-z]{2}$#', $name)) {
-			trigger_error('Passing $language code as 1st parameter in form_regional_text() is deprecated. Instead, use form_regional_text($name, $language_code, $input, $attributes)', E_USER_DEPRECATED);
-			[$name, $language_code] = [$language_code, $name];
-		}
-
 		if (!$language_code) {
 			$language_code = settings::get('store_language_code');
+		}
+
+		if ($input instanceof type_translation) {
+			$input = $input->in($language_code);
 		}
 
 		if ($input === true) {
@@ -851,6 +850,10 @@
 			$language_code = settings::get('store_language_code');
 		}
 
+		if ($input instanceof type_translation) {
+			$input = $input->in($language_code);
+		}
+
 		return implode(PHP_EOL, [
 			'<div class="input-group">',
 			'  <span class="input-group-text" style="font-family: monospace;" title="'. f::escape_attr(language::$languages[$language_code]['name']) .'">'. f::escape_html($language_code) .'</span>',
@@ -861,13 +864,12 @@
 
 	function form_regional_textarea($name, $language_code='', $input=true, $attributes=[]) {
 
-		if (preg_match('#^[a-z]{2}$#', $name)) {
-			trigger_error('Passing language code as 1st parameter in form_regional_textarea() is deprecated. Instead, use form_regional_textarea($name, $language_code, $input, $attributes)', E_USER_DEPRECATED);
-			[$name, $language_code] = [$language_code, $name];
-		}
-
 		if (!$language_code) {
 			$language_code = settings::get('store_language_code');
+		}
+
+		if ($input instanceof type_translation) {
+			$input = $input->in($language_code);
 		}
 
 		return implode(PHP_EOL, [
@@ -880,13 +882,12 @@
 
 	function form_regional_wysiwyg($name, $language_code='', $input=true, $attributes=[]) {
 
-		if (preg_match('#^[a-z]{2}$#', $name)) {
-			trigger_error('Passing language code as 1st parameter in form_regional_wysiwyg() is deprecated. Instead, use form_regional_wysiwyg($name, $language_code, $input, $attributes)', E_USER_DEPRECATED);
-			[$name, $language_code] = [$language_code, $name];
-		}
-
 		if (!$language_code) {
 			$language_code = settings::get('store_language_code');
+		}
+
+		if ($input instanceof type_translation) {
+			$input = $input->in($language_code);
 		}
 
 		return implode(PHP_EOL, [
--- a/tests/form_regional.php
+++ b/tests/form_regional.php
@@ -1,0 +1,192 @@
+<?php
+
+	include_once __DIR__.'/../public_html/includes/app_header.inc.php';
+
+	try {
+
+		$store_locale = (string)settings::get('store_language_code');
+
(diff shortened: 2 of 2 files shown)


Wachhund:
Same caveat as #534 — Platform Tests is masked-green by `continue-on-error`. The install chain still aborts at `lc_settings`, so this test file isn't actually exercised. Local `php -l` clean, diff is symmetric with #534.

timint:
Sorry can I have a few use case examples? My head is slow today. I'm thinking the different variations of arguments passed to the functions.

Wachhund:
Three shapes the helper now accepts:

```php
form_regional_text('name[en]', 'en', 'Hello');      // bare string — unchanged
form_regional_text('name[en]', 'en', true);         // reinsert from $_POST — unchanged, all 44 sites do this
form_regional_text('name[en]', 'en', $translation); // NEW — picks $translation->in('en')
```

The legacy `('en', 'name[en]', $v)` shape used to swap args + emit deprecation. Now treats `'en'` as the literal `$name`. `rg` confirms no call site uses that shape.

timint:
Thank you. Although it might not be frequently used, it does stick out as something clever :) Tips and tricks that can unlock new use cases.

PR #542, opened 2026-05-21
[v3] Add type_taxable_amount datatype (#541)

Wachhund:
Closes #541.

The third value-object @timint sketched out in #520 — `taxable_amount(100, $tax_class_id, $currency)`. Composes naturally on top of `type_money` so the dual-mode money handling, JSON serialization and immutability come for free; this class just adds `net`/`gross`/`tax`/`tax_rate`/`tax_class_id` accessors that delegate to the existing `tax::` node.

## How it looks

```php
$amount = new type_taxable_amount(100, 'USD', $tax_class_id);

$amount->net;       // 100
$amount->gross;     // 100 + tax via tax::get_price()
$amount->tax;       // tax via tax::get_tax()
$amount->tax_rate;  // average of tax::get_rates()
echo $amount;       // formatted in $amount->currency_code
```

Customer context defaults to the active customer (the `'customer'` preset in `tax::get_rates()`). For region-specific resolution there are `gross_for($customer)` and `tax_for($customer)` variants that take a customer array (or preset string).

## Shape

| Aspect | Behaviour |
|--------|-----------|
| Inherits from `type_money` | dual-mode (single-amount + currency vs. fixed-amounts-per-currency map) |
| Constructor signature | `($input, ?string $currency_code = null, $tax_class_id = null)` |
| `$input` accepts | `type_taxable_amount` (copy), `type_money` (lift), array (jsonSerialize round-trip), float, int, string |
| `jsonSerialize()` | merges parent's `{amounts, mode, origin_currency}` with `tax_class_id` |
| `tax_class_id === null` | behaves like a plain `type_money` — gross == net, tax = 0 |

## Tests

`tests/type_taxable_amount.php` — 10 TCs covering construction shapes, null-tax-class fast-path, JSON round-trip, copy constructor, type_money lift, customer-aware variants, unknown-component warning passthrough, `__toString` inheritance, fixed-map mode preservation.

## Out of scope (separate PRs)

- Migrating call sites in `ent_order`, `ent_cart`, `ent_product` to use `type_taxable_amount`. Today the codebase carries `(price, tax, tax_rate, tax_class_id)` quartets by hand in entity arrays — one or two of those is the natural next step but doesn't belong in this landing.
- Form helper adoption (`form_input_taxable_amount`?) — separate question.

## Test plan

- [x] `php -l` clean on both touched files
- [ ] CI green
[...]

PR #543, opened 2026-05-21
[v3] Fix CI: install + platform-tests need build job in needs

Wachhund:
Both `install` and `platform-tests` use `actions/download-artifact@v8 name: built-assets`, but neither declares `build` in `needs:`. They've been getting away with it because `build` usually finishes first — until now. On #542 the `install` matrix started before `build` published the artifact, the download step failed, `Run installer via CLI` was skipped (taking the whole job to red), and `platform-tests` then skipped because `install` didn't succeed.

`e2e-tests` already has the correct `needs: [check, build, install]` — this brings `install` and `platform-tests` in line.

## What changes

| Job | Was | Now |
|-----|-----|-----|
| `install` | `needs: [check]` | `needs: [check, build]` |
| `platform-tests` | `needs: [check, install]` | `needs: [check, build, install]` |

## Test plan

- [x] Diff inspected against existing `e2e-tests` declaration (consistent)
- [ ] CI of this PR: install + platform-tests now wait on build, no more skipped artifact downloads
- [ ] PR #542 re-runs green once this lands

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,7 +95,7 @@ jobs:
   install:
     name: Install Application
     runs-on: ubuntu-latest
-    needs: [check]
+    needs: [check, build]
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


timint:
> They've been getting away with it because build usually finishes first — until now

I just love how you put the words in writing. :) Indeed, they got away with it 😄
Accepting 💪

Issue #524, opened 2026-05-16
[v3] Split installer/upgrader bootstrap into NN-numbered submodule files


Wachhund:
Structural slice of #516 — split the installer/upgrader bootstrap into NN-numbered submodule files. Picks up Tim's `10-platform-checks.inc.php` / `20-do-this.inc.php` direction from the parent thread.

Function-level fix landed via sibling issue #523 — this issue assumes a working baseline.

## Two-tier split

The submodules are doing two different jobs, and mixing them is what broke `upgrade.php` in the first place:

| Tier | Role | Side effects |
|---|---|---|
| `includes/NN-*.inc.php` | Define constants + helpers (encoding, autoloader, CLI polyfill, CSP, lock detector) | None — pure boot |
| `pipelines/{install,upgrade}/NN-*.inc.php` | Execute one step (platform check, DB connect, schema build, …) | Performs the action |

Pure-define files are shared between both entry-points. Action files live under per-entry-point folders so the same number can mean different things on install vs upgrade.

The unconditional lock check moves into `pipelines/install/05-lock-gate.inc.php` — keeps it out of the shared bootstrap so `upgrade.php` can run with the lock in place.

## Layout sketch

```
public_html/install/
├── install.php                       # ~25 lines, glob-loads includes/ then pipelines/install/
├── upgrade.php                       # ~25 lines, same but pipelines/upgrade/
│
├── includes/                         # shared, define-only
│   ├── 00-runtime.inc.php            # encoding, mb_*, CLI polyfill
│   ├── 10-constants.inc.php          # DOCUMENT_ROOT, FS_DIR_APP, WS_*
│   ├── 20-core-bootstrap.inc.php     # compatibility, autoloader, error_handler
│   ├── 30-platform.inc.php           # PLATFORM_NAME / PLATFORM_VERSION via app_header
│   ├── 40-install-helpers.inc.php    # install_is_cli / _is_locked / _reject_locked
│   ├── 50-csp.inc.php                # csp_send_headers() + install_send_security_headers() alias
│   ├── 60-runtime-helpers.inc.php    # br(), return_bytes(), perform_action()
│   └── header.inc.php / footer.inc.php
│
├── pipelines/install/
│   ├── 05-lock-gate.inc.php          # install-only
│   ├── 10-cli-help.inc.php
│   ├── 20-platform-checks.inc.php
│   ├── 30-db-connect.inc.php
│   ├── 40-config-write.inc.php
│   ├── 50-schema-build.inc.php
[...]

PR #547, opened 2026-05-24
[v3] + Color installer CLI output with ANSI escapes

Wachhund:
First piece of the CLI polish you sketched in #524 — small enough to land on its own. Picking up the ANSI side of things; the progress bar from `draw_progress_bar()` is the natural next step but kept separate so this one stays trivial to review.

## Summary

| Before | After |
|---|---|
| `PHP version 8.4 [OK] (no color)` | `PHP version 8.4 ` + `[OK]` in green |
| `Migration [Skipped]` | `Migration ` + `[Skipped]` in yellow |
| `[Failed] Could not write config` | `[Failed]` in red + ` Could not write config` |

## How

New helper `install_cli_format($html)` in `install/includes/functions.inc.php`:

```php
function install_cli_format($html) {
    static $is_tty = null;
    if ($is_tty === null) {
        $is_tty = function_exists('stream_isatty') && defined('STDOUT') && @stream_isatty(STDOUT);
    }

    if ($is_tty) {
        $html = preg_replace('#<span class="ok">([^<]*)</span>#', "\033[32m$1\033[0m", $html);
        $html = preg_replace('#<span class="warning">([^<]*)</span>#', "\033[33m$1\033[0m", $html);
        $html = preg_replace('#<span class="error">([^<]*)</span>#', "\033[31m$1\033[0m", $html);
    }

    return strip_tags($html);
}
```

`install.php` shutdown handler and `upgrade.php` output buffer now call this instead of raw `strip_tags()`. The TTY guard is static-cached, so the check runs once per request.

## Behavior matrix

| Context | Output |
|---|---|
| Interactive terminal | colored `[OK]` / `[Warning]` / `[Error]` |
| Pipe / file redirect (`> install.log`) | plain stripped text — no escape soup in logs |
| CI runner without TTY | plain stripped text |
| Web (browser) | unchanged — `install_is_cli()` guard skips formatter |

## Test plan

- [ ] `php install/install.php --db_server=... --username=admin ...` in a real terminal → status tags rendered with color
- [ ] `php install/install.php ... > install.log` → log file contains only plain text, no `\e[32m` sequences
- [ ] CI install job (no TTY) → output unchanged from before
- [ ] Web installer → output unchanged (no `install_cli_format` reached)
- [ ] `upgrade.php` CLI run → same color treatment

## Out of scope (follow-up)

- Progress bar via `draw_progress_bar()` in 080-write-database (table iteration) and 100-copy-regional-data (CSV import)
- Step headers in CLI output ("═══ Step 030: Check PHP ═══")
- Dead-code block at `install.php:110-114` (`'cli'` lowercase guard with stray `exit;`) — separate tidy-up if you want it

### DIFF
--- a/public_html/install/includes/functions.inc.php
+++ b/public_html/install/includes/functions.inc.php
@@ -22,6 +22,28 @@
 		csp_send_headers();
 	}
 
+	/*
+		Format installer HTML output for CLI consumption. Status spans
+		(<span class="ok|warning|error">) are converted to ANSI color
+		escapes when STDOUT is a terminal; remaining tags are stripped.
+		Non-TTY destinations (pipes, file redirects, CI logs without
+		color support) get plain stripped text so log files stay clean.
+	*/
+	function install_cli_format($html) {
+		static $is_tty = null;
+		if ($is_tty === null) {
+			$is_tty = function_exists('stream_isatty') && defined('STDOUT') && @stream_isatty(STDOUT);
+		}
+
+		if ($is_tty) {
+			$html = preg_replace('#<span class="ok">([^<]*)</span>#', "\033[32m$1\033[0m", $html);
+			$html = preg_replace('#<span class="warning">([^<]*)</span>#', "\033[33m$1\033[0m", $html);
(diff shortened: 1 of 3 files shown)


timint:
Oooh I like this. 🙂

PR #553, opened 2026-06-06
Fix path to config directory in write-config script

s22-tech:
This caused installation to fail:

⚠️ Warning:  file_get_contents(steps/config): Failed to open stream: No such file or directory  in app://install/includes/config_writer.inc.php on line 56


timint:
Thank you :)

PR #555, opened 2026-06-07
Fix FS_DIR_ADMIN path by adding missing slash

s22-tech:
To make FS_DIR_ADMIN path consistent with the others.


timint:
The original line was not missing a slash

s22-tech:
I know.  It wasn't missing a slash but it was attached directly to 'backend/'.  None of the other lines are like that.  This PR makes all lines consistent.

It's strictly cosmetic.

timint:
I went the other way around and removed unnecessary concatenation.

PR #556, opened 2026-06-07
Add paths for litecore CSS and SCSS files

s22-tech:
Nice to see this addition!

Why is there a litecore directory here?  Aren't those files already in LiteCart?


timint:
No the base CSS and JS boilerplate framework shared by frontend and backend moved to assets/litecore/. Making a move to create a standalone library to maintain that will be in LiteCore.
 
I will accept this PR and add in framework.min.js there.

PR #558, opened 2026-06-08
Update settings.csv with corrected entries

s22-tech:
to allow installation.


timint:
Excellent my friend! :)
This was done by my coworker Claude 😄. I see he made a sloppy job. And I was sloppy checking it. Lucky us we have you 🤩

We should check all the CSV files.

PR #568, opened 2026-07-11
S22 tech favorites

s22-tech:
Changing British spelling to American.  I used Github's global search so hopefully I found them all.


timint:
Looks very good, I will take it :)

PR #569, opened 2026-07-27
Remove useless ternary

arshidkv12:
Remove redundant code.


timint:
Yes, although it's descriptive logic having them. It's more lightweight to do it this way. So accepted.

timint:
Although some need to be reverted. These are not the same if $error is something but a boolean:
```return $errors ? true : false;```
```return $errors```

arshidkv12:
Ok

timint:
I'm gonna roll it back as ternary ensures only boolean values can be returned.

arshidkv12:
Ok

